PR #1, opened 2020-02-04
Fix __init__.py.

dfhahn:
## Description
There is an import of a submodule called `freeenergyframework.read`, which is not available. I have deleted the line.

## Status
- [ ] Ready to go


hannahbrucemacdonald:
Good spot! That's a hangover from the molssi cookiecutter

PR #2, opened 2020-02-04
Suggest changes in plotting script.

dfhahn:
## Description
These changes in are merely cosmetic and include:
- change to quadratic plot
- fixing strings (i.e. superscript)
- change plot and font size
- changes color of data points dependent on deviation

## Questions
- Other color map?
- Cosmetics (line sizes, font sizes)?
- Different styles for dG and ddG plots for easier discrimination?
- All DDG plot: discriminate between simulated and not simulated edges?

## Status
- [] Ready to go

### DIFF
--- a/freeenergyframework/plotting.py
+++ b/freeenergyframework/plotting.py
@@ -40,8 +40,12 @@ def _master_plot(x, y, title='',
                          [ax_min + small_dist * 2, ax_max + small_dist * 2],
                          color='grey', alpha=0.2)
     # actual plotting
-    plt.scatter(x, y, color='hotpink')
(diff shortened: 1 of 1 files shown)


hannahbrucemacdonald:
LTGM!


I'd also like to have yours/de Groot's lab's colouring of datapoints according to distance to equality line... do you have that line of code to hand?

If not I can figure out how to write it!

dfhahn:
I have added the coloring of the data points dependent on the deviation. The error bars are gray, to color them it is a bit more tricky (`plt.errorbar` does not accept a list/array as argument for `color`)

hannahbrucemacdonald:
Thanks! I'm happy to leave them as grey for now, if we really want coloured error bars then we can loop over datapoints and add one scatter-point at a time, but lets not bother

PR #4, opened 2020-02-18
Add code to plotly plots

dfhahn:
## Description
This PR adds code to create correlation and barplots with `plotly`. The code is in a new submodule `plotlying`, which includes one function for barplots (`plot_bar`) and a new version of the function `_master_plot`. The functions in `plotting` have a keyword which lets the user choose between `maplotlib` or `plotly` plots.

## Todos
Not obligatory, but would be nice add-ons:
  - [ ] let the user choose the coloring scheme in the correlation plots (i.e. with a metric indicating the quality of a simulation (convergence).)
  - [ ] add more descriptive hovertext to the data points (i.e. names of ligands/ smile strings/ pictures of ligands) 

## Status
- [ ] Ready to go


hannahbrucemacdonald:
This looks good!


Do you think we could move the turning it into a dataframe into the `plot_bar()` function? The way it's written right now, this would only generate plots that are the raw DDG's from the edge, if we could pass in the FEMap.DiGraph object, and then turn it into a DF inside the function then it would be flexible to plot DG's or DDG's calculated from MLE?

Issue #3, opened 2020-02-04
RMSE/AUE error calculation


dfhahn:
There are different ways how the error on the RMSE, AUE is calculated in this `freeenergyframework` and `pmx`. The `freeenergyframework` method
draws bootstrap samples from the original sample, which does not account
for errors of the sample. The `pmx` method takes the datapoints from the original sample, and draws
the bootstrap sample's data point by drawing from a normal distribution
with center and standard deviation of the original sample's data point. Which is the correct way? 

See the example below:

```python
import numpy as np
from sklearn import metrics
import scipy
```

# Example data
Two sets of delta G values with associated errors

```python
# fake data
DG1  = np.random.random(100) * 10.0 - 5.0
dDG1 = np.random.normal(0, 1, size = 100)
DG2  = DG1 + np.random.normal(0, 2, size = 100)
dDG2 = np.random.normal(0, 1, size = 100)
```

```python
from matplotlib import pyplot as plt
plt.figure(figsize=(6,6))
plt.errorbar(DG1, DG2, xerr=dDG1, yerr=dDG2, ls='', marker='o')
plt.xlabel('DG1')
plt.ylabel('DG2')
plt.xlim(-7,7)
plt.ylim(-7,7)
```

    (-7, 7)

![RMSE Error Calculation_3_1](https://user-images.githubusercontent.com/8383063/73762817-ead5a280-4770-11ea-90b0-c087243838f3.png)

# `freeenergyframework` method
Take bootstrap samples from the actual sample, which does not account for errors of the sample

```python
nbootstrap = 1000
ci = 0.95
assert len(DG1) == len(DG2)
sample_size = len(DG1)
s_n = np.zeros([nbootstrap], np.float64) # s_n[n] is the statistic computed for bootstrap sample n
for replicate in range(nbootstrap):
    indices = np.random.choice(np.arange(sample_size), size=[sample_size], replace=True)
    DG1_sample, DG2_sample = DG1[indices], DG2[indices]
    s_n[replicate] = np.sqrt(metrics.mean_squared_error(DG1_sample, DG2_sample))
    
rmse_stats = dict()
rmse_stats['mle'] = np.sqrt(metrics.mean_squared_error(DG1, DG2))
rmse_stats['stderr'] = np.std(s_n)
rmse_stats['mean'] = np.mean(s_n)
# TODO: Is there a canned method to do this?
s_n = np.sort(s_n)
low_frac = (1.0-ci)/2.0
[...]

PR #5, opened 2020-02-19
Account for errors in statistics calculation (Issue #3).

dfhahn:
## Description
Fix issue #3 

## Todos
  - [ ] Account for errors in experimental and predicted samples when calculating the statistics

## Questions
- [ ] Bootstrap not based on perturbations (edges) but on ligands (nodes). Is this done already correctly?

## Status
- [ ] Ready to go

### DIFF
--- a/freeenergyframework/stats.py
+++ b/freeenergyframework/stats.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-def bootstrap_statistic(y_true, y_pred, ci=0.95, statistic='RMSE', nbootstrap = 1000, plot_type='dG'):
+def bootstrap_statistic(y_true, y_pred, dy_true=None, dy_pred=None, ci=0.95, statistic='RMSE', nbootstrap = 1000, plot_type='dG'):
     import sklearn.metrics
     import scipy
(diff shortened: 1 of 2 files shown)


hannahbrucemacdonald:
This looks good to me.

I see what you mean about bootstrapping based on edges, rather than nodes. The way this is currently done is correct for the `plot_all_DDGs` but not `plot_DDGs`. This function should do the same irrespective of the plot --- just take x, y, xerr and yerr.

I can make this change today!

freeenergyframework/stats.py
```diff
@@ -69,15 +73,27 @@ def unique_differences(x):
         N = len(x)
         return np.array( [ (x[i] - x[j]) for i in range(N) for j in range(N) if (i != j) ] )
 
+    if dy_true is None:
+        dy_true = np.zeros_like(y_true)
+    if dy_pred is None:
+        dy_pred = np.zeros_like(y_pred)
     assert len(y_true) == len(y_pred)
```
hannahbrucemacdonald: this could be

`assert len(y_true) == len(y_pred) == len(dy_true) == len(dy_pred)`

Issue #16, opened 2020-09-01
Allow for node names


hannahbrucemacdonald:
Some of the code `stats.mle` or specifically `stats.form_edge_matrix` expects the nodes of the graph to have integer names, and assumes that these are range(0,len(nodes)), which may not always be true. Some other node specifications might be preferable, such as ligand name or SMILES

PR #17, opened 2020-09-01
Allowing for non-integer node names

hannahbrucemacdonald:
Creates a dictionary for nodes, so that node names can be non-integers 
and non-contiguous


hannahbrucemacdonald:
Closes #16

PR #22, opened 2020-09-27
Allow for computations where both directions of an edge have been calculated

hannahbrucemacdonald:
## Description
form_edge_matrix now checks if the backwards edge has been run, and if so, uses a combination rule to use both to generate the matrix.

the matrix returned will be fully symmetric/antisymmetric

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/arsenic/stats.py
+++ b/arsenic/stats.py
@@ -217,11 +217,26 @@ def form_edge_matrix(g, label, step=None, action=None, node_label=None):
     for a, b in g.edges:
         i = node_name_to_index[a]
         j = node_name_to_index[b]
-        matrix[i, j] = g.edges[a, b][label]
+        if (b, a) in g.edges():
+            if action == 'symmetrize':
+                combination = 'geometric'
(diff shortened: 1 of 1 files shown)


jchodera:
Awesome!

But shouldn't it combine these in an uncertainty-aware manner? The [optimal way to combine two independent measurements](https://en.wikipedia.org/wiki/Inverse-variance_weighting) `A1, A2` with standard errors `dA1, dA2` is via the MLE:
```
A = w1 * A1 + w2*A2
dA = sqrt(1 / (dA1**(-2) + dA2**(-2))
w1 = dA1**(-2) / (dA1**(-2) + dA2**(-2))
w2 = dA2**(-2) / (dA1**(-2) + dA2**(-2))
```
The measurement with the smallest uncertainty will have greater weight.

jthorton:
implemented in #211

Issue #24, opened 2020-10-05
plot_DGs unusual result


callumjd:
Using the provided example.csv data, I cannot reproduce the DGs plot - the experimental data seems to be the negative of the true value.

Please let me know if I've done something wrong - PDF of notebook attached. Everything else looks correct.

[absolute_wrong.pdf](https://github.com/openforcefield/Arsenic/files/5330188/absolute_wrong.pdf)

hannahbrucemacdonald:
Hmmm I’ve recently made a change in sign to try keep things consistent, but I might not have tested it with all plot types so this looks like a bug. I will look into it and write a test so that it doesn’t happen again.


Sorry I’m still trying to figure out the right API so the package is still very much WIP!

callumjd:
Does seem like a sign issue - totally understand that it is WIP. Very nice for FEP analysis though, happy to be a beta tester!

hannahbrucemacdonald:
Ok

`import networkx as nx
g = nx.DiGraph()

for i,j in enumerate([-10,-9,-8]): 
    g.add_node(i, exp_DG=j, calc_DG=j,exp_dDG=0.2, calc_dDG=0.2)
    
plotting.plot_DGs(g)`


gives:
![image](https://user-images.githubusercontent.com/43652597/95682845-a2b02f00-0bdf-11eb-8c85-57def69caa52.png)


so it's not in the `plot_DGs` function... must be before that

hannahbrucemacdonald:
This should be fixed in master now! Sorry for slowness and thanks for flagging the mistake!

callumjd:
> This should be fixed in master now! Sorry for slowness and thanks for flagging the mistake!

Thanks! Have just tested, looks good 👍

PR #26, opened 2020-10-12
flipping sign of matrix to match expected

hannahbrucemacdonald:
## Description

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Arsenic
 ==============================
 [//]: # (Badges)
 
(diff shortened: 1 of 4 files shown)


hannahbrucemacdonald:
Addressing #24 

I flipped the sign in the MLE estimator before, so that node variables are returned with the same sign as inputted, however I then wrote the tests with the format `A minus B` for edge differences, when everywhere else is `B minus A` to redress this, the signs of the antisymmetric pairwise matrix.

This would have been caught with better testing, which I will add

PR #32, opened 2021-09-29
Fix for codecov

mikemhenry:
## Description
Codecov shows coverage at ~1% instead of ~48% so we have issue with our config


glass-w:
lgtm!

PR #34, opened 2021-11-24
Perses simulation results support

ijpulidos:
## Description
The main idea is supporting perses simulations results in arsenic. Creating also basic infrastructure for allowing ways to handle different formats other than the somewhat rigid csv.

## Todos
  - [ ] More testing with real life perses results

## Status
- [ ] Ready to go


ijpulidos:
We decided that this isn't the desired approach. I'm closing this and will be making a new one. We don't want perses-specific functionality.

PR #35, opened 2021-12-03
FEMap from object or dictionary

ijpulidos:
## Description
This allows creating an `FEMap` from either a dictionary or a the newly created object `FEResults` which stores absolute experimental and relative calculated free energy results.

## Todos
  - [x] Allow a pythonic and programmatic way of creating an FEMap. Other than using a static CSV file.


openff/arsenic/arsenic.py
```diff
@@ -1,5 +1,5 @@
 # this will combine a single command that will run all analysis and save everything
-from . import plotting, wrangle
+from openff.arsenic import plotting, wrangle
```
mikemhenry, line 2: Why the change here?

openff/arsenic/wrangle.py
```diff
@@ -59,12 +61,109 @@ class ExperimentalResult(object):
...
+        Construct Free Energy map of simulations from input data.
+
+        Parameters
+        ----------
+            input_data: csv file path or path-like object or dict-like object
+                File path to csv file or python dict-like object.
+
+        Examples
+        --------
+        To read from a csv file specifically formatted for this, you can use:
```
mikemhenry, line 134: Maybe mention here (or link to it if it is in the docs somewhere else) what format the CSV file needs to be in.

mikemhenry (changes requested):
Can you add a test?  Then this should be good to go! Other points were just a few questions/suggestions.

openff/arsenic/arsenic.py
ijpulidos: The PEP8 [recommendation](https://www.python.org/dev/peps/pep-0008/#imports) is to use absolute imports, but I guess both should work.

mikemhenry: That's what I thought, let's use absolute imports!

openff/arsenic/wrangle.py
ijpulidos: As far as I can see this format isn't documented. We just have the `data/examples.csv` file that I'm referencing in the example code. I think we could link to that here, I agree. Thanks!

jchodera: We shouldn't link to the file---we should figure out what it is expecting and explicitly specify it here.

jchodera (commented):
This is looking much better!

PR #38, opened 2022-01-08
Fix #37: Fix erroneous MLE estimate when self-edges are included

### LINKED ISSUE #37: Graphs containing self-edges will produce erroneous MLE estimates
It appears that graphs provided to `arsenic.stats.mle` can have self-edges (edges that connect to the same node at both ends).

[This line](https://github.com/openforcefield/openff-arsenic/blob/master/openff/arsenic/stats.py#L191) assumes that cannot be the case, and will produce biased, incorrect estimates if provided with self-edges.

We should either automatically omit self-edges or raise an exception if they are encountered.

jchodera:
## Description
This PR fixes issue #37 (Graphs containing self-edges will produce erroneous MLE estimates), and adds a test that failed prior to the fix to verify the issue is resolved.

## Status
- [x] Ready to go

### DIFF
--- a/openff/arsenic/stats.py
+++ b/openff/arsenic/stats.py
@@ -146,6 +146,8 @@ def mle(
     Xu, Huafengraph. "Optimal measurement network of pairwise differences."
     Journal of Chemical Information and Modeling 59.11 (2019): 4720-4728.
 
+    NOTE: Self-edges (edges that connect a node to itself) will be ignored.
+
     Parameters
     ----------
     graph :nx.Graph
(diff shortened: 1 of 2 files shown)


openff/arsenic/stats.py
```diff
@@ -209,6 +214,9 @@ def mle(
     for (a, b) in graph.edges:
         i = node_name_to_index[a]
         j = node_name_to_index[b]
+        if i == j:
+            # The MLE solver will fail if we include self-edges, so we need to omit these
+            continue
```
ijpulidos, line 219: Nice catch! This looks good. Just thinking that the loops in lines 193 and 214 could be merged (should increase performance). But that would also require that the loop in line 210 to happen before. It's probably a small increase in performance but thought it should be mentioned. Non-blocking.

ijpulidos (approved):
Nice catch. I have some non-blocking comments, but this should be merged as it is.

PR #60, opened 2022-05-17
Fix CI Badge

mikemhenry:


mikemhenry:
@IAlibay is there a reason we don't just run black to fix these errors? Or just add a pre-commit?

IAlibay:
> @IAlibay is there a reason we don't just run black to fix these errors? Or just add a pre-commit? 

Richard kinda merged things eagerly before I could ask in the previous PR 🤣

I would say just fix the errors now and we can go pre-commit later?

PR #64, opened 2022-09-01
Improve plotting.py

zhang-ivy:
## Description
This PR improves plotting functions in plotting.py. It does not alter any functions in plotlying.py

## Todos
  - [x] Fix formatting of x and y axis labels
  - [x] Add argument (`dpi`) for controlling the resolution of the plot when saving
  - [x] Allow each edge to be labeled using the names of the nodes involved in the edge. This enhancement was only introduced for DDG plots. Handles both small molecules and protein mutations. Does not work for plotly-generated plots.
  - [x] Use the `adjustText` package to adjust data labels so that they are not overlapping
  - [x] Make the minimum and maximum axes of the plot larger

## Questions
- [ ] Should we merge this into the main repo or should I just merge it in my fork since some of the changes may be specific to my needs I'm working on?

## Status
- [ ] Ready to go

### DIFF
--- a/cinnabar/plotting.py
+++ b/cinnabar/plotting.py
@@ -27,6 +27,7 @@ def _master_plot(
     shift: float = 0.0,
     figsize: float = 3.25,
     dpi: float = 'figure',
+    data_labels: list = [],
 ):
     """Handles the aesthetics of the plots in one place.
 
@@ -74,6 +75,8 @@ def _master_plot(
         the resolution in dots per inch
         if 'figure', uses the figure's dpi value (this behavior is copied from
         https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html)
+     data_labels : list of str, default []
+        list of labels for each data point
 
     Returns
     -------
@@ -141,6 +144,10 @@ def _master_plot(
     )
     plt.scatter(x, y, color=color, s=10, marker="o", zorder=2)
 
+    # Label points
+    for i, label in enumerate(data_labels):
+        plt.text(x[i] + .03, y[i] + .03, label, fontsize=9)
+
     # stats and title
     statistics_string = ""
     for statistic in statistics:
@@ -196,6 +203,17 @@ def plot_DDGs(
     symmetrise : bool, default = False
         whether to plot each datapoint twice, both
         positive and negative
+    plotly : bool, default = False
+        whether to use plotly to generate the plot
+    data_label_type : str or None, default = None
+        type of data label to add to each edge
+        if None, data labels will not be added
+        if 'small-molecule', edge labels will be f"{node_A_name}-{node_B_name}"
+        if 'protein-mutation', edge labels will be f"{node_A_name}{node_B_name[0]}"
+            where node names are formatted f"{one_letter_amino_acid_code}{residue id}"
+            e.g. if node A is Y29 and node B is A29, the edge label will be Y29A
+        currently unsupported for plotly-generated plots
+        TODO: implement data labeling for the case where plotly=True
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


zhang-ivy:
@mikemhenry @ijpulidos Can you review this and help me determine whether this a PR worth merging into the main repo or if I should just merge it into my own fork? Some of the changes I made may be specific to my own needs, though I did try to make the changes as general as possible. Specifically, the following changes may be useful for others, but I implemented them only for the case that I'm interested in: plotting DDGs (and not DGs) using plotting.py (and not plotly.py):
- Allow each edge to be labeled using the names of the nodes involved in the edge. This enhancement was only introduced for DDG plots. Handles both small molecules and protein mutations. Does not work for plotly-generated plots.
- Use the adjustText package to adjust data labels so that they are not overlapping

cinnabar/plotting.py
```diff
@@ -83,11 +92,11 @@ def _master_plot(
...
 
-    plt.xlabel(f"{xlabel} {quantity} / " + units)
-    plt.ylabel(f"{ylabel} {quantity} / " + units)
+    plt.xlabel(f"{xlabel} {quantity} ({units})")
+    plt.ylabel(f"{ylabel} {quantity} ({units})")
 
-    ax_min = min(min(x), min(y)) - 0.5
-    ax_max = max(max(x), max(y)) + 0.5
+    ax_min = min(min(x), min(y)) - 1.0
+    ax_max = max(max(x), max(y)) + 1.0
```
mikemhenry, line 99: What is this change for?

zhang-ivy: In the PR description, its: 
Make the minimum and maximum axes of the plot larger

zhang-ivy: In my plots, this change is necessary because otherwise some of the error bars are cut off, so I extended the plot to make it bigger

zhang-ivy: The more general fix for this is probably to adjust `ax_min` and `ax_max` to find the min and max of the error bars, not free energies.

mikemhenry: This is the only change (as far as I can tell) that changes behavior in a way where a plot made with this PR would look different than a plot made with the current code. How about you add another argument called `error_padding` or something that defaults to `0.5` but then a user can pass in whatever value they want.

zhang-ivy:
@mikemhenry : What do you think about including `adjustText` as a dependency -- are you ok with it? This PR would require it.

mikemhenry:
@zhang-ivy thank you for pointing that out! I'm fine with it, given that 
https://github.com/conda-forge/adjusttext-feedstock/blob/main/recipe/meta.yaml
it doesn't pull in anything extra and is on conda-forge! (I typed this out so that later I remember why we decided to do this)

mikemhenry:
TODO 
- [ ] add `adjustText` to conda env yaml

cinnabar/plotting.py
```diff
@@ -201,13 +228,32 @@ def plot_DDGs(
...
     y = [x[2]["calc_DDG"] for x in graph.edges(data=True)]
+    xerr = np.asarray([x[2]["exp_dDDG"] for x in graph.edges(data=True)])
+    yerr = np.asarray([x[2]["calc_dDDG"] for x in graph.edges(data=True)])
+
+    # labels
+    data_labels = []
+    if data_label_type:
+        node_names = {node_id:  node_data["name"] for node_id, node_data in graph.nodes(data=True)}
+        if data_label_type == "small-molecule":
+            data_labels = [f"{node_names[node_A]}-{node_names[node_B]}" for node_A, node_B, data in graph.edges(data=True)]
```
zhang-ivy, line 245: Use unicode arrow instead of `-`

zhang-ivy: This is done!

cinnabar/plotting.py
zhang-ivy: This is done!

zhang-ivy:
@mikemhenry : I think I'm done making changes to this! Could you add `adjustText` to the conda env yaml?

mikemhenry (approved):

PR #69, opened 2022-12-01
Some fixes

annamherz:
Hello! I've been using cinnabar and have noticed some small fixes:

[1.](https://github.com/annamherz/cinnabar/blob/20fa6cb8b98bf56265e87eea51e4be8db811d3b9/cinnabar/arsenic.py#L42) If this is not `network.graph` it isn't able to plot.

[2.](https://github.com/annamherz/cinnabar/blob/20fa6cb8b98bf56265e87eea51e4be8db811d3b9/cinnabar/plotting.py#L422) Similarly here, without `shift` the function does not work. I've added it in but I guess it could also be removed from [the end](https://github.com/annamherz/cinnabar/blob/20fa6cb8b98bf56265e87eea51e4be8db811d3b9/cinnabar/plotting.py#L501) instead for it to be able to proceed.

[3.] `self.n_edges = len(self.results)` always returns two, as the length of the dictionary returned by the `read_csv` function is two. I think [this here](https://github.com/annamherz/cinnabar/blob/20fa6cb8b98bf56265e87eea51e4be8db811d3b9/cinnabar/wrangle.py#L108) would be better to get the actual number of edges in the network?

Thank you!

### DIFF
--- a/cinnabar/arsenic.py
+++ b/cinnabar/arsenic.py
@@ -39,8 +39,8 @@ if __name__ == "__main__":
     if "network" in args.plot:
         network.draw_graph(title=args.title, filename=f"{args.prefix}network.png")
     if "ddg" in args.plot:
-        plotting.plot_DDGs(network.results, title=args.title, filename=f"{args.prefix}DDGs.png")
+        plotting.plot_DDGs(network.graph, title=args.title, filename=f"{args.prefix}DDGs.png")
     if "dg" in args.plot:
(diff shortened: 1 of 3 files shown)


richardjgowers:
hi @annamherz thanks, these are three good fixes!

richardjgowers:
hi @annamherz thanks, these are three good fixes!

richardjgowers:
@annamherz this is in the v0.2.3 release of cinnabar that happened today on conda-forge.  It also includes `cinnabar` being installed as a command line tool, so you should be able to use it that way now.

PR #72, opened 2022-12-03
Rework FEMap object

richardjgowers:
## Description
Complete rework of FEMap object.

Changes
- now `FEMap.__init__` takes no arguments
- use `FEMap.add_measurement()` to add data to the graph
- either `AbsoluteMeasurement` or `RelativeMeasurement` are `openff.models` to be passed in
- use `FEMap.from_csv` to create from legacy csv file
- added `FEMap.to_legacy_graph()` to create the old `.graph` attribute
- `FEMap.graph` now a `nx.MultiDiGraph` (allowing multiple edges between nodes)
- each estimate/measurement of free energy difference is an independent edge (not lots of values stacked per edge)
- cli usage should remain unchanged (for now)


cinnabar/wrangle.py
```diff
@@ -22,22 +23,28 @@ def read_csv(filepath: pathlib.Path) -> dict:
...
+                                      float(other_err))
                 raw_results["Calculated"].append(calc)
     return raw_results
 
 
-class RelativeResultFromCSV(object):
-    def __init__(self, ligandA, ligandB, calc_DDG, mbar_error, other_error):
-        self.ligandA = str(ligandA).strip()
-        self.ligandB = str(ligandB).strip()
+class RelativeResult:
```
richardjgowers, line 39: Rather than `RelativeResult` and `ExperimentalResult` (and eventually `AbsoluteResult`) maybe it makes more sense to just have `Relative` and `Absolute` where the experimental/computational-ness of an object is just another field?

cinnabar/wrangle.py
```diff
@@ -58,85 +65,35 @@ def toDF(self):
...
-            ligand, expt_value, expt_error
-        )
 
-    def add_calculated_result(self, ligand_a, ligand_b, calc_value, mbar_error, other_error):
-        """
-        Add calculated relative result from plain explicit values.
+class FEMap:
+    results: list[Union[RelativeResult, ExperimentalResult]]
+    computational_graph: nx.MultiDiGraph
+    experimental_graph: nx.MultiDiGraph
```
richardjgowers, line 78: This seemed like the nicest approach... So two graphs, one is bench measurements, one computational measurements.  Both Multi graphs as you could have multiple measurements of the same node/edge (todo: not sure if NX allows multiple identical nodes (i.e. multiple absolute measurements))

cinnabar/wrangle.py
```diff
@@ -58,85 +65,35 @@ def toDF(self):
...
-                File path to csv file or python dict-like object.
+            input_data: dict-like object
+                dict-like object.
 
         Examples
         --------
         To read from a csv file specifically formatted for this, you can use:
 
-        >>> fe = wrangle.FEMap('../data/example.csv')
+        >>> fe = wrangle.FEMap.from_csv('../data/example.csv')
```
richardjgowers, line 96: I think we just need this break where we make this constructor a classmethod, but it already doesn't handle absolute computational results so maybe it'd be kinder to just kill it

cinnabar/wrangle.py
richardjgowers: so a way to do multiple absolute measurements is to have **all** measurements as edges, and absolute measurements are from a dummy node (I used `"NULL"`)

cinnabar/wrangle.py
```diff
@@ -22,209 +23,137 @@ def read_csv(filepath: pathlib.Path) -> dict:
                 expt_block = False
                 calc_block = True
             if expt_block and len(line.split(",")) == 3 and line[0] != "#":
-                expt = ExperimentalResultFromCSV(*line.split(","))
+                ligand, DG, dDG = line.split(",")
+                expt = AbsoluteMeasurement(ligand=ligand,
+                                           DF=float(DG),
+                                           uncertainty=float(dDG),
+                                           computational=False)
```
IAlibay, line 30: Rather than computational, maybe calculated or estimated?

cinnabar/wrangle.py
```diff
@@ -22,209 +23,137 @@ def read_csv(filepath: pathlib.Path) -> dict:
                 expt_block = False
                 calc_block = True
             if expt_block and len(line.split(",")) == 3 and line[0] != "#":
-                expt = ExperimentalResultFromCSV(*line.split(","))
+                ligand, DG, dDG = line.split(",")
+                expt = AbsoluteMeasurement(ligand=ligand,
+                                           DF=float(DG),
```
IAlibay, line 28: why DF over DG when using DDG elsewhere?

cinnabar/wrangle.py
IAlibay: Would it make sense (maybe longer term) to just have arbitrary numbers of graphs here?

I.e. if I have results from ITC, NMR, etc... and then some calculated metrics, I'd like to just dump them all in the same class and do all my analyses at the same time?

cinnabar/wrangle.py
```diff
@@ -22,209 +23,137 @@ def read_csv(filepath: pathlib.Path) -> dict:
...
-        self.n_edges = self.graph.number_of_edges()
-        self.degree = self.n_edges / self.n_ligands
-
-        # check the graph has minimal connectivity
-        self.check_weakly_connected()
-        if not self.weakly_connected:
-            print("Graph is not connected enough to compute absolute values")
+        # unpack data dictionary
+        fe = cls()
+        for r in data['Calculated']:
```
IAlibay, line 93: to save my sanity, do we really need to have capitalization here and for "Experimental"?

cinnabar/wrangle.py
```diff
@@ -22,209 +23,137 @@ def read_csv(filepath: pathlib.Path) -> dict:
...
+    def n_ligands(self) -> int:
+        """Total number of unique ligands"""
+        return len(self.experimental_graph.nodes | self.computational_graph.nodes)
+
+    @property
+    def degree(self) -> float:
+        """Average degree of all nodes"""
+        return self.n_measurements / self.n_ligands
+
+    def is_weakly_connected(self) -> bool:
```
IAlibay, line 147: long term we should consider calling this just "is_connected" given that's exactly what does, you can't really tell if it's weakly or strongly connected?

IAlibay (dismissed):
Looks like a great improvement, could comments / requested changes, mostly sanity things.

Tests I assume might come later?

cinnabar/wrangle.py
richardjgowers: this is legacy stuff, I'm not improving it (and it'll get removed eventually)

cinnabar/wrangle.py
richardjgowers: could do...  I think I'm going to allow arbitrary annotations per measurement (which end up as edge annotations on the graph), this is where e.g. `mbar_err` can live.  Then on the experimental graph you could have `source="ITC"` on some measurements.

Though that's also an argument you could have one graph, with edges first annotated as `type='computational'`

richardjgowers:
@IAlibay re tests: the plan is to push up tests for the existing code, then continually rebase this on top of those new tests.  Test coverage currently doesn't include any plotting functionality, but is otherwise ok.

cinnabar/wrangle.py
IAlibay: you know what I mean..

cinnabar/wrangle.py
IAlibay: I thought we were boldly changing stuff, but yeah fair I can't wait :P

cinnabar/wrangle.py
IAlibay: > Though that's also an argument you could have one graph, with edges first annotated as type='computational'

Pretty much :P 

Also technically we don't need to treat computational vs experimental differently, we just need to know what we're using as our ground truth and what we're adjusting, so you could just go "label A is experiment, label B is calculated", etc...

cinnabar/wrangle.py
```diff
@@ -41,18 +42,18 @@ def read_csv(filepath: pathlib.Path) -> dict:
...
+    uncertainty: FloatQuantity['kilojoule_per_mole']
     computational: bool
 
 
-class AbsoluteMeasurement(BaseModel):
+class AbsoluteMeasurement(DefaultModel):
     ligand: str
-    DG: float
-    uncertainty: float
+    DG: FloatQuantity['kilojoule_per_mole']
```
mikemhenry, line 55: The nice thing about this is currently with the way openff models works is if a users passes in kcal/mol, it will get converted automatically

cinnabar/femap.py
```diff
@@ -0,0 +1,177 @@
...
+    def n_measurements(self) -> int:
+        """Total number of both experimental and computational measurements"""
+        return len(self.experimental_graph.edges) + len(self.computational_graph.edges)
+
+    @property
+    def n_ligands(self) -> int:
+        """Total number of unique ligands"""
+        # must ignore NULL sentinel node
+        exptl = self.experimental_graph.nodes - {'NULL'}
+        compt = self.computational_graph.nodes
```
IAlibay, line 140: what if you have non overlapping comp & experimental nodes?

IAlibay: i.e. do we have checks in place somewhere to avoid that edge case?

cinnabar/femap.py
```diff
@@ -0,0 +1,177 @@
...
+        """Average degree of all nodes"""
+        return self.n_measurements / self.n_ligands
+
+    def is_weakly_connected(self) -> bool:
+        """Checks if all results in the graph are reachable from other results"""
+        # todo; cache
+        undirected_graph = self.computational_graph.to_undirected()
+        return nx.is_connected(undirected_graph)
+
+    def generate_absolute_values(self):
```
IAlibay, line 155: to-do post merge: we should try to fix this to work on cases where there are fewer experimental nodes than computational ones, it's a big ask by partners

cinnabar/femap.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Any other attributes on the measurement are used as annotations
+
+        Raises
+        ------
+        ValueError : if bad type given
+        """
+        if isinstance(measurement, AbsoluteMeasurement):
+            # coerce to relative to simplify logic
+            meas_ = RelativeMeasurement(labelA='NULL',
```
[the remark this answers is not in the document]
IAlibay: Also this brings about some subsequent queries about the science - if you have an absolute measurement and you turn it to relative, what happens when you apply an MLE to it? Do you derive an absolute on an absolute?

cinnabar/femap.py
IAlibay: Actually there's nothing to _fix_ here I think. The MLE with will MLE, and should work irrespective of the number of associated experimental nodes. What we probably should methodify is the experimental shift, i.e. https://github.com/OpenFreeEnergy/cinnabar/blob/0f10e11129d5fd2589893794398e528d406b6581/cinnabar/plotting.py#L422-L424

Then folks could call to access shifted dGs if necessary.

cinnabar/femap.py
```diff
@@ -0,0 +1,177 @@
...
+        elif isinstance(measurement, RelativeMeasurement):
+            meas_ = measurement
+        else:
+            raise ValueError()
+
+        # slurp out tasty data, anything but labels
+        d = dict(meas_)
+        d.pop('labelA', None)
+        d.pop('labelB', None)
+        d.pop('label', None)
```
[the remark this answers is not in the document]
richardjgowers: labels are nodes.  This code is transforming a measurement `{'labelA': 'foo', 'labelB': 'bar', 'DG': 1.0, 'uncertainty': 0.1}` into an edge, so it strips out the label info and keeps anything else as annotation on the edge

cinnabar/femap.py
```diff
@@ -0,0 +1,177 @@
...
+        else:
+            self.experimental_graph.add_edge(meas_.labelA, meas_.labelB, **d)
+
+    @property
+    def n_measurements(self) -> int:
+        """Total number of both experimental and computational measurements"""
+        return len(self.experimental_graph.edges) + len(self.computational_graph.edges)
+
+    @property
+    def n_ligands(self) -> int:
```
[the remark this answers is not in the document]
richardjgowers: state maybe?  nodes?  but it's also confusing if this refers to comp, expt, or both types of nodes.  previously comp and expt had the same answer, but we're breaking that assumption

cinnabar/femap.py
richardjgowers: see above, I don't know how `n_nodes` is defined for non-overlapping cases.

cinnabar/femap.py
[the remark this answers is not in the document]
IAlibay: This is where the separation of graphs is kinda difficult and we probably should avoid it in future iterations. I guess for now you could split this into n_comp_nodes and n_exp_nodes?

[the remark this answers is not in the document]
richardjgowers: @IAlibay there's a guy called Irfan asking me over here to allow fewer experimental nodes than comp, can you two agree on if you want matched graphs or not: https://github.com/OpenFreeEnergy/cinnabar/pull/72#discussion_r1126628435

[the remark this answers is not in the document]
IAlibay: I'm not sure why you couldn't have a single graph with both comp and exp and also not enforce that the edges / nodes for both types need to overlap? This needs you to have the "source" tag in there, but then you have a set of nodes with edges, some of which might be experimental and some of which might not?

richardjgowers:
@IAlibay ok it's all one graph now, see if you prefer this

cinnabar/femap.py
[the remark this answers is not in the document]
IAlibay: Questions on this were answered yesterday I believe.

cinnabar/femap.py
```diff
@@ -0,0 +1,235 @@
...
+            comp_graph.add_edge(a, b)
+
+        return nx.is_connected(comp_graph)
+
+    def generate_absolute_values(self):
+        # TODO: Make this return a new Graph with computational nodes annotated with DG values
+
+        # TODO this could work if either relative or absolute expt values are provided
+        if self.check_weakly_connected():
+            graph = self.to_legacy_graph()
```
IAlibay, line 167: I assume this is a temporary solution?

IAlibay (approved):
lgtm!

cinnabar/femap.py
richardjgowers: yes.  as you've raised, things like MLE will need some rethinking in the light of absolute values, so for now `to_legacy_graph` works as before (and drops abs values)

Issue #73, opened 2023-01-16
95% CI sometimes doesn't contain the RMSE or MUE


zhang-ivy:
I generated a cinnabar plot for my terminally-blocked amino acid RBFE calc data, comparing the forward vs reverse DDGs, but the RMSE and MUE do not lie within their respective 95% CIs.

![image (19)](https://user-images.githubusercontent.com/35546250/212754333-cd4fe26e-6362-430c-ac54-7a2901b03f2b.png)

Here is the code that does the bootstrapping: https://github.com/OpenFreeEnergy/cinnabar/blob/1dbe248511e179b49d7cb654d22b46ebbd15fd46/cinnabar/stats.py#L111-L130
![image (19)](https://user-images.githubusercontent.com/35546250/212753909-b8c03db9-14d2-42e1-9e66-9f1732d22ceb.png)

@jchodera points out that  "The bootstrapping is both (1) resampling with replacement and (2) adding noise related to the experimental error to the experimental data and noise related to the predicted error in each replicate.
If we intend to capture the finite sample size effect, we only want (1), and do not want to include (2).
Can you try eliminating the extra addition of normal error in lines 117 and 118? You can just say "scale=0*..." to set the scale to zero by just adding the 0* in place rather than rewriting those lines."

When I re-generate the plot with John's suggested fix, the problem goes away: 

![image](https://user-images.githubusercontent.com/35546250/212754648-d1156eb0-3b21-4fd2-a970-bac0a6262e5d.png)

IAlibay:
So part of the "issue" here is that we're using the RMSE (same is true  based on a direct calculation of the RMSE from the x / y inputs (i.e. what's available under the 'mle' key) rather than the mean of the bootstrapped results (i.e. what's available under 'mean' key).

@jchodera are you saying that the noise in the experimental error shouldn't be included at all or that it shouldn't be included based on a given use case? If it's the latter, would it make sense to also report the estimate of the bootstrapped population as means of helping to assess how likely the original sample estimate represents the true value?

zhang-ivy:
From @jchodera: 
We should figure out whether the API should support a fixed version of the old mode ("simulate what would happen if I ran the experiment and calculation many times", where we need to report the mean of the bootstraps, not the MLE) in addition to the scale 0 mode you just implemented ("show me the effects of finite sample size so I can draw some conclusions about how well we fit the data", which I think should be the default)

zhang-ivy:
[...]

PR #74, opened 2023-01-18
Update how RMSE and MUE (and their CIs) are computed

zhang-ivy:
## Description
Fixes https://github.com/OpenFreeEnergy/cinnabar/issues/73 -- where in some cases, the RMSE and MUE does not lie within the 95% CI

## Todos
  - [x] Change the default bootstrapping mode to only account for finite sample size effects (tells us how well we fit the data)
  - [x] Support the old bootstrapping mode (which also accounts for simulating what would happen if we ran the experiment and/or calculation many times) by allowing the user to separately control whether experimental and predicted uncertainty should be included when bootstrapping

<strike>  - [x] For both modes, instead of reporting the MLE of RMSE and MUE, report the mean of the bootstrapped data</strike>
  - [x] For the default mode, continue to report the sample estimate of the RMSE and MUE. For the old mode, report the mean of the bootstrapped data

## Status
- [x] Ready to go

### DIFF
--- a/cinnabar/plotlying.py
+++ b/cinnabar/plotlying.py
@@ -274,8 +274,8 @@ def _master_plot(
                                         xerr,
                                         yerr,
                                         statistic=statistic,
-                                        bootstrap_true_uncertainty=bootstrap_x_uncertainty,
-                                        bootstrap_pred_uncertainty=bootstrap_y_uncertainty)
+                                        include_true_uncertainty=bootstrap_x_uncertainty,
+                                        include_pred_uncertainty=bootstrap_y_uncertainty)
 
         statistic_type = 'mean' if bootstrap_x_uncertainty or bootstrap_y_uncertainty else 'mle'
         string.append(
--- a/cinnabar/plotlying.py
+++ b/cinnabar/plotlying.py
@@ -160,6 +160,7 @@ def _master_plot(
     filename: Optional[str] = None,
     bootstrap_x_uncertainty: bool = False,
     bootstrap_y_uncertainty: bool = False,
+    statistic_type: str = "mle",
 ):
     nsamples = len(x)
     ax_min = min(min(x), min(y)) - 0.5
@@ -268,6 +269,8 @@ def _master_plot(
 
     # stats and title
     string = []
+    if statistic_type not in ['mle', 'mean']:
+        raise Exception(f"Unknown statistic type {statistic_type}")
     for statistic in statistics:
         bss = stats.bootstrap_statistic(x,
                                         y,
@@ -276,8 +279,6 @@ def _master_plot(
                                         statistic=statistic,
                                         include_true_uncertainty=bootstrap_x_uncertainty,
                                         include_pred_uncertainty=bootstrap_y_uncertainty)
-
-        statistic_type = 'mean' if bootstrap_x_uncertainty or bootstrap_y_uncertainty else 'mle'
         string.append(
             f"{statistic + ':':5s}{bss[statistic_type]:5.2f} [95%: {bss['low']:5.2f}, {bss['high']:5.2f}]"
         )
--- a/cinnabar/plotlying.py
+++ b/cinnabar/plotlying.py
@@ -277,8 +277,9 @@ def _master_plot(
                                         bootstrap_true_uncertainty=bootstrap_x_uncertainty,
                                         bootstrap_pred_uncertainty=bootstrap_y_uncertainty)
 
+        statistic_type = 'mean' if bootstrap_x_uncertainty or bootstrap_y_uncertainty else 'mle'
         string.append(
-            f"{statistic + ':':5s}{bss['mean']:5.2f} [95%: {bss['low']:5.2f}, {bss['high']:5.2f}]"
+            f"{statistic + ':':5s}{bss[statistic_type]:5.2f} [95%: {bss['low']:5.2f}, {bss['high']:5.2f}]"
         )
     stats_string = "<br>".join(string)
 
--- a/cinnabar/plotlying.py
+++ b/cinnabar/plotlying.py
@@ -267,10 +267,16 @@ def _master_plot(
     # stats and title
     string = []
     for statistic in statistics:
-        bss = stats.bootstrap_statistic(x, y, statistic=statistic)
+        bss = stats.bootstrap_statistic(x,
+                                        y,
+                                        xerr,
+                                        yerr,
+                                        statistic=statistic,
+                                        bootstrap_experimental_uncertainty=bootstrap_experimental_uncertainty,
+                                        bootstrap_predicted_uncertainty=bootstrap_predicted_uncertainty)
 
         string.append(
-            f"{statistic + ':':5s}{bss['mle']:5.2f} [95%: {bss['low']:5.2f}, {bss['high']:5.2f}]"
+            f"{statistic + ':':5s}{bss['mean']:5.2f} [95%: {bss['low']:5.2f}, {bss['high']:5.2f}]"
         )
     stats_string = "<br>".join(string)
 
--- a/cinnabar/plotlying.py
+++ b/cinnabar/plotlying.py
@@ -270,7 +270,7 @@ def _master_plot(
     # stats and title
     string = []
     if statistic_type not in ['mle', 'mean']:
-        raise Exception(f"Unknown statistic type {statistic_type}")
+        raise ValueError(f"Unknown statistic type {statistic_type}")
     for statistic in statistics:
         bss = stats.bootstrap_statistic(x,
                                         y,
--- a/cinnabar/plotlying.py
+++ b/cinnabar/plotlying.py
@@ -158,6 +158,8 @@ def _master_plot(
     origins: bool = True,
     statistics: list = ["RMSE", "MUE"],
     filename: Optional[str] = None,
+    bootstrap_x_uncertainty: bool = False,
+    bootstrap_y_uncertainty: bool = False,
 ):
     nsamples = len(x)
     ax_min = min(min(x), min(y)) - 0.5
@@ -272,8 +274,8 @@ def _master_plot(
                                         xerr,
                                         yerr,
                                         statistic=statistic,
-                                        bootstrap_experimental_uncertainty=bootstrap_experimental_uncertainty,
-                                        bootstrap_predicted_uncertainty=bootstrap_predicted_uncertainty)
+                                        bootstrap_true_uncertainty=bootstrap_x_uncertainty,
+                                        bootstrap_pred_uncertainty=bootstrap_y_uncertainty)
 
         string.append(
             f"{statistic + ':':5s}{bss['mean']:5.2f} [95%: {bss['low']:5.2f}, {bss['high']:5.2f}]"
--- a/cinnabar/plotting.py
+++ b/cinnabar/plotting.py
@@ -95,7 +95,8 @@ def _master_plot(
     bootstrap_y_uncertainty : bool, default False
         whether to account for uncertainty in y when bootstrapping
     statistic_type : str, default 'mle'
-        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
+        the type of statistic to use, either 'mle' (i.e. sample statistic)
+        or 'mean' (i.e. bootstrapped mean statistic)
 
     Returns
     -------
@@ -260,7 +261,8 @@ def plot_DDGs(
     bootstrap_y_uncertainty : bool, default False
         whether to account for uncertainty in y when bootstrapping
     statistic_type : str, default 'mle'
-        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
+        the type of statistic to use, either 'mle' (i.e. sample statistic)
+        or 'mean' (i.e. bootstrapped mean statistic)
 
     Returns
     -------
@@ -398,7 +400,8 @@ def plot_DGs(
     bootstrap_y_uncertainty : bool, default False
         whether to account for uncertainty in y when bootstrapping
     statistic_type : str, default 'mle'
-        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
+        the type of statistic to use, either 'mle' (i.e. sample statistic)
+        or 'mean' (i.e. bootstrapped mean statistic)
 
     Returns
     -------
@@ -492,7 +495,8 @@ def plot_all_DDGs(
     bootstrap_y_uncertainty : bool, default False
         whether to account for uncertainty in y when bootstrapping
     statistic_type : str, default 'mle'
-        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
+        the type of statistic to use, either 'mle' (i.e. sample statistic)
+        or 'mean' (i.e. bootstrapped mean statistic)
 
     Returns
     -------
--- a/cinnabar/plotting.py
+++ b/cinnabar/plotting.py
@@ -34,6 +34,7 @@ def _master_plot(
     font_sizes: dict = {"title": 12, "labels": 9, "other": 12},
     bootstrap_x_uncertainty: bool = False,
     bootstrap_y_uncertainty: bool = False,
+    statistic_type: str = "mle",
 ):
     """Handles the aesthetics of the plots in one place.
 
@@ -93,6 +94,8 @@ def _master_plot(
         whether to account for uncertainty in x when bootstrapping
     bootstrap_y_uncertainty : bool, default False
         whether to account for uncertainty in y when bootstrapping
+    statistic_type : str, default 'mle'
+        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
     Returns
     -------
 
@@ -171,6 +174,8 @@ def _master_plot(
 
     # stats and title
     statistics_string = ""
+    if statistic_type not in ['mle', 'mean']:
+        raise Exception(f"Unknown statistic type {statistic_type}")
     for statistic in statistics:
         s = stats.bootstrap_statistic(x,
                                       y,
@@ -179,7 +184,6 @@ def _master_plot(
                                       statistic=statistic,
                                       include_true_uncertainty=bootstrap_x_uncertainty,
                                       include_pred_uncertainty=bootstrap_y_uncertainty)
-        statistic_type = 'mean' if bootstrap_x_uncertainty or bootstrap_y_uncertainty else 'mle'
         string = f"{statistic}:   {s[statistic_type]:.2f} [95%: {s['low']:.2f}, {s['high']:.2f}] " + "\n"
         statistics_string += string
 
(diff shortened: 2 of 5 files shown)


zhang-ivy:
@IAlibay : Done! Let me know if you think there is a better test to add

cinnabar/tests/test_stats.py
```diff
@@ -145,3 +145,24 @@ def test_correlation_positive(fe_map):
         assert (
             0.5 < bss["mle"] < 0.9
         ), f"Correlation must be positive for this data. {stat} is {bss['mle']}"
+
```
IAlibay, line 148:
```suggestion


```
PEP8

cinnabar/tests/test_stats.py
```diff
@@ -145,3 +145,24 @@ def test_correlation_positive(fe_map):
...
+def test_confidence_intervals(fe_map):
+    """
+    Test that the confidence intervals for RMSE and MUE contain the corresponding statistics.
+    Uses the example data in `cinnabar/data/example.csv`
```
IAlibay, line 152:
```suggestion
    Test that the confidence intervals for RMSE and MUE contain the
    corresponding statistics. Uses the example data in
    `cinnabar/data/example.csv`
```
PEP8

cinnabar/tests/test_stats.py
```diff
@@ -145,3 +145,24 @@ def test_correlation_positive(fe_map):
...
+    nodes = fe_map.graph.nodes
+
+    x_data = np.asarray([n[1]["exp_DG"] for n in nodes(data=True)])
+    y_data = np.asarray([n[1]["calc_DG"] for n in nodes(data=True)])
+    xerr = np.asarray([n[1]["exp_dDG"] for n in nodes(data=True)])
+    yerr = np.asarray([n[1]["calc_dDG"] for n in nodes(data=True)])
+
+    # RMSE
+    bss = bootstrap_statistic(x_data, y_data, xerr, yerr, statistic="RMSE")
+    assert bss['low'] < bss['mean'] < bss['high'], "The RMSE must lie withint the bootstrapped 95% CI"
```
IAlibay, line 164: This is a good first test but it doesn't actually trigger either of the newly introduced bools for adding noise either in the x or y direction. Ideally we need a test for that, probably maybe a regression test (assuming enough bootstraps to ensure convergence)?

cinnabar/stats.py
```diff
@@ -14,6 +14,8 @@ def bootstrap_statistic(
     statistic: str = "RMSE",
     nbootstrap: int = 1000,
     plot_type: str = "dG",
+    bootstrap_experimental_uncertainty: bool = False,
```
IAlibay, line 17: I would prefer we use the labels `x_true` and `y_pred` for these in the same manner as the other arguments for the method. Whilst the normal use is experiment & predicted, there is no reason why this can't be predicted1 vs predicted2, or even experiment1 vs experiment2.

IAlibay (changes requested):
Sorry I don't think I will have time today to dig into what would make an ideal test for this, I can try to come back to it tomorrow if needed.

If possible could you check for other PEP8 issues in the newly added code please? Specifically lines 164 and 168 seem to exceed 79 chars.

cinnabar/stats.py
zhang-ivy: Fixed!

cinnabar/plotting.py
```diff
@@ -167,8 +172,14 @@ def _master_plot(
...
-        s = stats.bootstrap_statistic(x, y, xerr, yerr, statistic=statistic)
-        string = f"{statistic}:   {s['mle']:.2f} [95%: {s['low']:.2f}, {s['high']:.2f}] " + "\n"
+        s = stats.bootstrap_statistic(x,
+                                      y,
+                                      xerr,
+                                      yerr,
+                                      statistic=statistic,
+                                      bootstrap_experimental_uncertainty=bootstrap_experimental_uncertainty,
+                                      bootstrap_predicted_uncertainty=bootstrap_predicted_uncertainty)
+        string = f"{statistic}:   {s['mean']:.2f} [95%: {s['low']:.2f}, {s['high']:.2f}] " + "\n"
```
IAlibay, line 182: Just an additional thought here, by using 'mean' for all cases here we do slightly move away from what's in the original issue (I think?).

Now we're saying that in all cases the estimate from the bootstrap samples rather than the original sample estimate is the thing we report. Assuming a statistic that is a linear function[note1] of the original data and enough bootstraps, I assume the two will come close but they are unlikely to be the same?

My understanding is that folks preferred using the sample estimate where possible (i.e. the case of adding noise is one of those cases where it's not possible). I'm not sure I have a formed opinion yet on which is better to be the most reported value[note2], that being said I do believe it to be important that we at least warn users of the difference between the two estimates, ensuring that folks are aware of any bias in the estimates.

[note1]: Are we guaranteeing this?
[note2]: I think I take the stance that ideally both would be reported when communicating science, but I do know it gets messy.

zhang-ivy: >My understanding is that folks preferred using the sample estimate where possible (i.e. the case of adding noise is one of those cases where it's not possible).

I was unsure of this as well -- I thought @jchodera had mentioned in the call yesterday that in both the default and old modes, we want to use the mean instead of the MLE, but now that I'm thinking about it, he might have meant that when using either of the newly introduced bools are used, we want to use the mean. 

I will update the code so that in the default mode, the sample estimate (`s['MLE']`) is used. If one of the newly introduced bools is used, we used mean (`s['mean']`)

cinnabar/stats.py
ijpulidos: I agree with this suggestion, it is better if we have names that tell more. Maybe something that  has `true` and `predicted` on the names, `include_true_errors` and `include_predicted_errors` or something similar. By the name of the function we already know this is related to boostrapping, so it feels unnecessary to have it in the `kwarg` name.

cinnabar/plotting.py
zhang-ivy: Just pushed the changes

IAlibay: If @jchodera has the time to review this it would be grand, it is one of those things where it seems like an innocuous change, but I do worry it will have wider consequences than I can see.

Maybe an opinion by @ppxasjsm would be good? That way we're a bit more sure we aren't messing up other's pipelines?

cinnabar/stats.py
zhang-ivy: Just changed `bootstrap` to `include` for the arguments to `bootstrap_statistics`. Did not change the argument names plotting.py and plotlying.py functions

cinnabar/tests/test_stats.py
zhang-ivy: I just changed the test to use the newly introduced bools and check whether the RMSE lies within the 95% CIs. 

I thought about adding a regression test -- was thinking I could add my data from #73 to `cinnabar/data/` and then create a test that checks whether that specific dataset passes the test, but I'm not using a standard cinnabar csv input file for that dataset, because its not the standard "predicted DDG" vs "experiment DDG", its "forward DDG" vs "reverse DDG." It would take a bit of time to properly format it and I'm not sure its essential to include, so I decided not to.

zhang-ivy: Let me know if you think of better tests / feel free to add them yourself if you have time!

cinnabar/plotting.py
```diff
@@ -167,8 +172,15 @@ def _master_plot(
...
-        s = stats.bootstrap_statistic(x, y, xerr, yerr, statistic=statistic)
-        string = f"{statistic}:   {s['mle']:.2f} [95%: {s['low']:.2f}, {s['high']:.2f}] " + "\n"
+        s = stats.bootstrap_statistic(x,
+                                      y,
+                                      xerr,
+                                      yerr,
+                                      statistic=statistic,
+                                      include_true_uncertainty=bootstrap_x_uncertainty,
+                                      include_pred_uncertainty=bootstrap_y_uncertainty)
+        statistic_type = 'mean' if bootstrap_x_uncertainty or bootstrap_y_uncertainty else 'mle'
```
jchodera, line 182: Instead of auto-selecting the mean or mle depending on whether bootstrap_{x|y}_uncertainty is specified, can we just have a separate kwarg for `statistic_type` that defaults to `mle` and allows the user to set it to `mean` if they prefer? That would make the behavior much more obvious.

jchodera (changes requested):
There's one minor change that would help reduce the surprise developers will have in using this, but other than this, it looks great!

zhang-ivy:
@IAlibay @ijpulidos : could you review this again / merge if it looks good?

cinnabar/plotting.py
```diff
@@ -342,6 +373,9 @@ def plot_DGs(
     plotly: bool = False,
     centralizing: bool = True,
     shift: float = 0.0,
+    bootstrap_x_uncertainty: bool = False,
+    bootstrap_y_uncertainty: bool=False,
```
ijpulidos, line 377: Just minor style fix.
```suggestion
    bootstrap_y_uncertainty: bool = False,
```

ijpulidos (approved):
Looks good to me! Just a minor style change. Not blocker.

cinnabar/plotting.py
zhang-ivy: Good catch, thanks!

cinnabar/plotting.py
```diff
@@ -87,7 +90,12 @@ def _master_plot(
...
     font_sizes : dict, default {"title": 12, "labels": 9, "other": 12}
         font sizes to use for the title ("title"), the data labels ("labels"), and the rest of the plot ("other")
-
+    bootstrap_x_uncertainty : bool, default False
+        whether to account for uncertainty in x when bootstrapping
+    bootstrap_y_uncertainty : bool, default False
+        whether to account for uncertainty in y when bootstrapping
+    statistic_type : str, default 'mle'
+        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
     Returns
```
IAlibay, line 99:
```suggestion

    Returns
```

cinnabar/plotlying.py
```diff
@@ -266,11 +269,18 @@ def _master_plot(
 
     # stats and title
     string = []
+    if statistic_type not in ['mle', 'mean']:
+        raise Exception(f"Unknown statistic type {statistic_type}")
```
IAlibay, line 273:
```suggestion
        raise ValueError(f"Unknown statistic type {statistic_type}")
```
Please don't raise a plain Exception, use a `ValueError` here? Also please add test for exception.

cinnabar/plotting.py
```diff
@@ -166,9 +174,17 @@ def _master_plot(
 
     # stats and title
     statistics_string = ""
+    if statistic_type not in ['mle', 'mean']:
+        raise Exception(f"Unknown statistic type {statistic_type}")
```
IAlibay, line 178:
```suggestion
        raise ValueError(f"Unknown statistic type {statistic_type}")
```
As above + test

cinnabar/plotting.py
```diff
@@ -358,6 +392,12 @@ def plot_DGs(
         Title for the plot
     filename : str, default = None
         filename for plot
+    bootstrap_x_uncertainty : bool, default False
+        whether to account for uncertainty in x when bootstrapping
+    bootstrap_y_uncertainty : bool, default False
+        whether to account for uncertainty in y when bootstrapping
+    statistic_type : str, default 'mle'
+        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
```
IAlibay, line 400: As above

IAlibay (changes requested):

cinnabar/plotting.py
```diff
@@ -87,7 +90,12 @@ def _master_plot(
         contains the minimium and maximum values to use for the x and y axes. if specified, axis_padding is ignored
     font_sizes : dict, default {"title": 12, "labels": 9, "other": 12}
         font sizes to use for the title ("title"), the data labels ("labels"), and the rest of the plot ("other")
-
+    bootstrap_x_uncertainty : bool, default False
+        whether to account for uncertainty in x when bootstrapping
+    bootstrap_y_uncertainty : bool, default False
+        whether to account for uncertainty in y when bootstrapping
+    statistic_type : str, default 'mle'
+        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
```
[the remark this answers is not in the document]
zhang-ivy: Done

cinnabar/plotting.py
```diff
@@ -235,6 +254,12 @@ def plot_DDGs(
             the negative of the transformation
         currently unsupported for plotly-generated plots
         TODO: implement data labeling for the case where plotly=True
+    bootstrap_x_uncertainty : bool, default False
+        whether to account for uncertainty in x when bootstrapping
+    bootstrap_y_uncertainty : bool, default False
+        whether to account for uncertainty in y when bootstrapping
+    statistic_type : str, default 'mle'
+        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
```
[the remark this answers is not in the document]
zhang-ivy: Done

cinnabar/plotting.py
zhang-ivy: Done

cinnabar/plotting.py
```diff
@@ -437,6 +486,12 @@ def plot_all_DDGs(
         whether to use plotly for the plotting
     shift : float, default = 0.
         shift both the x and y axis by a constant
+    bootstrap_x_uncertainty : bool, default False
+        whether to account for uncertainty in x when bootstrapping
+    bootstrap_y_uncertainty : bool, default False
+        whether to account for uncertainty in y when bootstrapping
+    statistic_type : str, default 'mle'
+        the type of statistic to use, either 'mle' or 'mean' (of the bootstrapped data)
```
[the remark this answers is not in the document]
zhang-ivy: Done

cinnabar/tests/test_stats.py
```diff
@@ -145,3 +145,47 @@ def test_correlation_positive(fe_map):
...
+    nodes = fe_map.graph.nodes
+
+    x_data = np.asarray([n[1]["exp_DG"] for n in nodes(data=True)])
+    y_data = np.asarray([n[1]["calc_DG"] for n in nodes(data=True)])
+    xerr = np.asarray([n[1]["exp_dDG"] for n in nodes(data=True)])
+    yerr = np.asarray([n[1]["calc_dDG"] for n in nodes(data=True)])
+
+    # RMSE (default mode)
+    bss = bootstrap_statistic(x_data, y_data, xerr, yerr, statistic="RMSE")
+    assert bss['low'] < bss['mle'] < bss['high'], \
```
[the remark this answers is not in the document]
zhang-ivy: Done

cinnabar/tests/test_stats.py
```diff
@@ -145,3 +145,47 @@ def test_correlation_positive(fe_map):
...
+        "The RMSE must lie within the bootstrapped 95% CI"
+
+    # MUE (default mode)
+    bss = bootstrap_statistic(x_data, y_data, xerr, yerr, statistic="MUE")
```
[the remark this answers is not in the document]
zhang-ivy: @IAlibay : I'm not familiar with `pytest.parametrize` -- would you have time to re-format this with `pytest.parametrize`? If not, maybe we could just merge this PR as is and I could open an issue about improving the tests in the long term? While I'm sure your suggestion would improve the code, I'm not sure its essential, as my goal here was just to introduce the minimal changes necessary so that I (and others in the Chodera lab) can continue making progress on the scientific side of things.

cinnabar/tests/test_stats.py
```diff
@@ -145,3 +145,47 @@ def test_correlation_positive(fe_map):
...
+        "The RMSE must lie within the bootstrapped 95% CI"
+
+    # RMSE (including yerr in bootstrapping)
+    bss = bootstrap_statistic(x_data, y_data, xerr, yerr, statistic="RMSE",
+                              include_pred_uncertainty=True)
+    assert bss['low'] < bss['mean'] < bss['high'], \
+        "The RMSE must lie within the bootstrapped 95% CI"
+
+    # RMSE (including both xerr and yerr in bootstrapping)
+    bss = bootstrap_statistic(x_data, y_data, xerr, yerr, statistic="RMSE",
```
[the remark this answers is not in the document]
zhang-ivy: > We're not testing passing statistic_type through
@IAlibay : That's because `statistic_type` is not an argument that's passed to `bootstrap_statistic`, it is only passed through `plotting.py` functions. We currently don't test any `plotting.py` functions, so would it be ok if I just opened an issue about writing tests for `plotting.py` as a long term action item?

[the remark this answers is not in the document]
zhang-ivy: > Is our dataset such that these could ever fail? Can we add the dataset from the problematic example (i.e. https://github.com/OpenFreeEnergy/cinnabar/issues/73 ) to make sure we test against that edge case properly?

I added a test addressing the case in my latest commit. I hard coded the value directly into the test (which @dotsdl mentioned would be ok), since I was having trouble formatting the cinnabar input CSV to properly handle my dataset (previously described this problem here: https://github.com/OpenFreeEnergy/cinnabar/pull/74#discussion_r1073863391)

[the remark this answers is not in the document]
IAlibay: >  We currently don't test any plotting.py functions, so would it be ok if I just opened an issue about writing tests for plotting.py as a long term action item?

Sorry that's on me, I should have realized that. Happy to go with just raising an issue about testing the plotting methods.

IAlibay (approved):

cinnabar/tests/test_stats.py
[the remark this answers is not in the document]
IAlibay: Done

cinnabar/tests/test_stats.py
```diff
@@ -145,3 +146,71 @@ def test_correlation_positive(fe_map):
...
+    x_data = [-0.101, 0.351, 0.117, 0.623, 5.172, 5.209, -1.727, -1.387, -1.534, 1.082]
+    y_data = [-0.174, 0.42, 0.262, 0.626, 5.064, 4.783, -1.58, -1.712, -1.699, 0.822]
+    xerr = [0.443, 0.652, 0.57, 0.245, 1.112, 1.049, 1.23, 1.435, 1.521, 0.505]
+    yerr = [0.442, 0.714, 0.619, 0.224, 1.401, 1.107, 1.178, 1.252, 1.265, 0.472]
+
+    # RMSE (default mode)
+    bss = bootstrap_statistic(x_data, y_data, xerr, yerr, statistic="RMSE",
+                              include_true_uncertainty=True,
+                              include_pred_uncertainty=True)
+    assert (bss['low'] > bss['mle']) or (bss['mle'] > bss['high']), error_message
```
zhang-ivy, line 216: Was reviewing @IAlibay 's changes and just realized that `error_message` isn't defined here. I'll open a PR to fix this.

zhang-ivy:
Thanks for your help and feedback @IAlibay ! Will try to incorporate pytest fixtures / parametrize in the future

cinnabar/tests/test_stats.py
zhang-ivy: Here's the PR: https://github.com/OpenFreeEnergy/cinnabar/pull/76
Note that I also found another issue with the edge case test and fixed it in the above PR

PR #76, opened 2023-01-25
Fix test_confidence_interval_edge_case

zhang-ivy:
## Description
Fixes `test_confidence_interval_edge_case` which was introduced in https://github.com/OpenFreeEnergy/cinnabar/pull/74

## Todos
  - [x] Defines `error_message` (previously undefined)
  - [x] Fixed the assert statement, which was previously checking that when using the old behavior (i.e. use the sample statistic and include the true and predicted uncertainty when bootstrapping), the sample RMSE statistic lies outside of the 95% confidence interval. We instead want to test that the new default behavior works properly for the old data.

## Status
- [x] Ready to go

### DIFF
--- a/cinnabar/tests/test_stats.py
+++ b/cinnabar/tests/test_stats.py
@@ -214,4 +214,4 @@ def test_confidence_interval_edge_case():
                               include_true_uncertainty=False,
                               include_pred_uncertainty=False)
     error_message = "The stat must lie within the bootstrapped 95% CI"
-    assert (bss['low'] < bss['mle']) or (bss['mle'] < bss['high']), error_message
+    assert (bss['low'] < bss['mle']) and (bss['mle'] < bss['high']), error_message
--- a/cinnabar/tests/test_stats.py
+++ b/cinnabar/tests/test_stats.py
@@ -211,6 +211,7 @@ def test_confidence_interval_edge_case():
 
     # RMSE (default mode)
     bss = bootstrap_statistic(x_data, y_data, xerr, yerr, statistic="RMSE",
-                              include_true_uncertainty=True,
-                              include_pred_uncertainty=True)
-    assert (bss['low'] > bss['mle']) or (bss['mle'] > bss['high']), error_message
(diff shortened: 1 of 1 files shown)


cinnabar/tests/test_stats.py
```diff
@@ -211,6 +211,7 @@ def test_confidence_interval_edge_case():
 
     # RMSE (default mode)
     bss = bootstrap_statistic(x_data, y_data, xerr, yerr, statistic="RMSE",
-                              include_true_uncertainty=True,
-                              include_pred_uncertainty=True)
-    assert (bss['low'] > bss['mle']) or (bss['mle'] > bss['high']), error_message
+                              include_true_uncertainty=False,
```
IAlibay, line 214: I think I'm missing something obvious here - I'm not sure I understand why CI was passing with these being set to True. It sounds like this test (which should reproduce the behaviour of the original) issue isn't working as expected?

IAlibay (changes requested):
changes lgtm but I'm going to temporarily block on some introspection on why CI didn't catch this

cinnabar/tests/test_stats.py
```diff
@@ -211,6 +211,7 @@ def test_confidence_interval_edge_case():
...
+                              include_pred_uncertainty=False)
+    error_message = "The stat must lie within the bootstrapped 95% CI"
+    assert (bss['low'] < bss['mle']) or (bss['mle'] < bss['high']), error_message
```
IAlibay, line 217: Please disregard my previous comment, I guess the issue was the comparison not being accurate - also shouldn't this be `and` instead of `or`?

IAlibay (changes requested):

cinnabar/tests/test_stats.py
zhang-ivy: Yes good catch!

zhang-ivy: Fixed

IAlibay (approved):

PR #78, opened 2023-01-25
Allow control of marker size (and other aspects of plt.scatter)

zhang-ivy:
## Description
Allow user to control the size of the markers (and other arguments) in plt.scatter()

## Todos
  - [x] Introduces `scatter_kwargs` argument to `plotting._master_plot`

## Status
- [x] Ready to go

### DIFF
--- a/cinnabar/plotting.py
+++ b/cinnabar/plotting.py
@@ -29,11 +29,10 @@ def _master_plot(
     figsize: float = 3.25,
     dpi: Union[float, str] = "figure",
     data_labels: list = [],
-    data_label_filter: Optional[float] = None,
     axis_padding: float = 0.5,
     xy_lim: list = [],
     font_sizes: dict = {"title": 12, "labels": 9, "other": 12},
-    marker_size: float = 10,
+    marker_size : float = 10,
 ):
     """Handles the aesthetics of the plots in one place.
 
@@ -83,8 +82,6 @@ def _master_plot(
         https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html)
     data_labels : list of str, default []
         list of labels for each data point
-    data_label_filter : float, default None
-        only show labels for data points where the abs(x - y) > data_label_filter
     axis_padding : float, default = 0.5
         padding to add to maximum axis value and subtract from the minimum axis value
     xy_lim : list, default []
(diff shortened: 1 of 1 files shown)


zhang-ivy:
@mikemhenry or @ijpulidos  : Could you review this?

cinnabar/plotting.py
```diff
@@ -97,6 +98,8 @@ def _master_plot(
     statistic_type : str, default 'mle'
         the type of statistic to use, either 'mle' (i.e. sample statistic)
         or 'mean' (i.e. bootstrapped mean statistic)
+    marker_size : float, default 10
```
IAlibay, line 101: I'm tempted to say that this point you really should be just passing in a dictionary of kwargs for matplotlib's scatter rather than incrementally adding new kwargs.

zhang-ivy: Good point. I just pushed changes to address this suggestion. Let me know what you think

cinnabar/plotting.py
```diff
@@ -35,6 +35,7 @@ def _master_plot(
     bootstrap_x_uncertainty: bool = False,
     bootstrap_y_uncertainty: bool = False,
     statistic_type: str = "mle",
+    scatter_kwargs: dict = {},
```
IAlibay, line 38: Would it be worth keeping s=10 and marker='o' as the defaults here? That way you have the same behaviour as before just with a little bit extra control?

IAlibay (changes requested):

cinnabar/plotting.py
zhang-ivy: Done!

cinnabar/plotting.py
```diff
@@ -97,6 +98,8 @@ def _master_plot(
     statistic_type : str, default 'mle'
         the type of statistic to use, either 'mle' (i.e. sample statistic)
         or 'mean' (i.e. bootstrapped mean statistic)
+    scatter_kwargs : dict, default {}
```
IAlibay, line 101: Could you update the default value here too please?

IAlibay (approved):
Just the one thing so I'll approve since it's just docs. @mikemhenry could you double check (sorry my brain is all mushy from this cold)?

cinnabar/plotting.py
zhang-ivy: Yes, sorry about that. Done

mikemhenry (approved):
:+1:

PR #81, opened 2023-03-08
Try to fix rtd

IAlibay:
Whilst I'm here, looks like the commented out import is breaking RTD, is this something we can undo now or it is WIP @richardjgowers ?


richardjgowers:
I think we need to look at making plotly optional (or maybe it's fine for it to be required)

PR #84, opened 2023-03-23
remove AbsoluteMeasurement concept

richardjgowers:
introduce GroundState object

instead all measurements are relative to a GroundState

## Description

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/cinnabar/femap.py
+++ b/cinnabar/femap.py
@@ -129,8 +129,14 @@ class FEMap:
     @property
     def n_ligands(self) -> int:
         """Total number of unique ligands"""
+        return len(self.ligands)
+
+    @property
+    def ligands(self) -> list:
+        """All ligands in the graph"""
         # must ignore ReferenceState nodes
-        return sum(1 for n in self.graph.nodes if not isinstance(n, ReferenceState))
+        return [n for n in self.graph.nodes
+                if not isinstance(n, ReferenceState)]
 
     @property
     def degree(self) -> float:
@@ -184,6 +190,33 @@ class FEMap:
                         source='MLE',
                     )
                 )
+
+            # find all computational result labels
+            comp_ligands = set()
+            for A, B, d in self.graph.edges(data=True):
+                if not d['computational']:
+                    continue
+                comp_ligands.add(A)
+                comp_ligands.add(B)
+
+            # find corresponding experimental results
+
+            # use mean of experimental results to offset MLE reference point
+
+            # add connection to MLE reference state and true reference state
+            self.add_measurement(
+                Measurement(
+                    labelA=ReferenceState(),
+                    labelB=g,
+                    DG=0.1*u,
+                    uncertainty=0.0 * u,
+                    computational=True,
+                    source='MLE',
+                )
+            )
+        else:
+            # TODO: This can eventually be worked around surely?
+            raise ValueError("Computational results are not fully connected")
 
     def to_legacy_graph(self) -> nx.DiGraph:
         """Produce single graph version of this FEMap
--- a/cinnabar/femap.py
+++ b/cinnabar/femap.py
@@ -4,7 +4,7 @@ from typing import Union
 import openff.units
 from openff.units import unit
 import warnings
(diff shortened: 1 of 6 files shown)


cinnabar/femap.py
```diff
@@ -25,6 +24,9 @@ def read_csv(filepath: pathlib.Path, units: Optional[openff.units.Quantity] = No
     raw_results = {"Experimental": {}, "Calculated": []}
     expt_block = False
     calc_block = False
+
+    ground = GroundState()
```
IAlibay, line 28: worth documenting that ground state is assumed to be the same for all inputs for this method?

IAlibay: Also is there a unique string we could give it so it hashes differently?

cinnabar/femap.py
```diff
@@ -65,19 +68,20 @@ class FEMap:
     >>> # Load/create experimental results
     >>> from openff.units import unit
     >>> kJpm = unit.kilojoule_per_mole
-    >>> experimental_result1 = AbsoluteMeasurement(label="CAT-13a", DG=-8.83 * kJpm, uncertainty=0.10 * kJpm)
-    >>> experimental_result2 = AbsoluteMeasurement(label="CAT-17g", DG=-9.73 * kJpm, uncertainty=0.10 * kJpm)
+    >>> g = GroundState()
```
IAlibay, line 71: Could you introduce the idea of label here so that folks can understand that it can be set to different states that hash differently?

IAlibay (approved):
Overall lgtm, I really like this idea.

Only comments are:
1. A bit more documentation on behaviour would be good
2. It's hard to tell if this is good because _I_ like it, I'd be ok with merging this quickly, but if we can have someone else gloss over the idea afterwards that'd be grand.

jchodera:
Is "Ground state" a reference state with absolute free energy defined to be zero? If so, maybe ZeroState or ReferenceState is more clear?

DiffNet permits multiple absolute free energy measurements of multiple states relative to that zero. Will it still be possible to have multiple absolute measurements?

richardjgowers:
@jchodera `GroundState` is *a* reference state, [`TrueGround`](https://github.com/OpenFreeEnergy/cinnabar/pull/84/files#diff-a17d12184142c333f7346f8262cc42f58f15acf08f85dd8d726637ea9c3deac9R50) is the theoretical zero point, and is a special case of GroundState.  "Absolute" measurements are against one of these two types of points.

For example, MLE would first place computed ligands against an arbitrary MLE `GroundState`, then next draws a measurement between that GroundState and the GroundState of the experimental (absolute) measurements.  The derived absolute computational values are then the sum of the edge to the arbitrary ground and the edge to the ground of the experimental points.


```mermaid
flowchart TD
  G[Experimental GroundState] -- experiment --> A[experimental A];
  G  -- experiment --> D[experimental B]; 
  H[MLE GroundState] --> B[computational A];
  H --> C[computational B];
  B -- RBFE --> C;
  G -- derived --> H;
  Z[True Ground] -- ??? --> G;
```

IAlibay:
Is there really a need for something like `TrueGround`? If I'm being only slightly cynical, the standard deviation around even high quality experiment is such that you are really just setting your ground state to a given set of experimental conditions at a given time in space.

richardjgowers:
this diagram is better actually:

```mermaid
flowchart TD
  G[Experimental GroundState] -- experiment --> A[ligand A];
  G  -- experiment --> B[ligand B]; 
  H[MLE GroundState] -- MLE --> A;
  H -- MLE --> B;
  A -- RBFE --> B;
  G -- derived --> H;
  Z[True Ground] -- assumed 0 --> G;
```

@IAlibay `TrueGround` is maybe unnecessary?  The way some sort of `FEMap.get_absolute('ligand A')` would work would be to pathfind from `TrueGround` to `ligand A`, which would be the sum of three edges (one of which is zero, so could collapse it away?).

PR #92, opened 2023-07-18
convert IC50 to DG

hannahbaumann:
## Description
Added function for converting IC50 into dG

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


cinnabar/measurements.py
```diff
@@ -52,19 +54,52 @@ class Measurement(DefaultModel):
 
     @classmethod
     def from_experiment(cls,
-                        pIC50: float,
+                        IC50: float,
```
richardjgowers, line 57: looking at the uncertainty about units below, maybe this, uncertainty and temperature should be given in `unit.Quantity` values.  There's some cool stuff you can do:

```python
In [1]: from openff.units import unit

In [2]: 5.0 * unit.micromolar
Out[2]: 5.0 <Unit('micromolar')>

In [3]: m = 5.0 * unit.micromolar

In [4]: m.is_compatible_with(unit.molar)
Out[4]: True

In [5]: m.is_compatible_with(unit.kelvin)
Out[5]: False

In [6]: m.to(unit.nanmolar)
Out[6]: 4999.999999999999 <Unit('nanomolar')>

```

cinnabar/measurements.py
```diff
@@ -52,19 +54,52 @@ class Measurement(DefaultModel):
...
+        uncertainty: float, optional
+          uncertainty of the experimental value in nM, default is zero if no uncertainty is provided
+        source: str, optional
+          source of experimental measurement
+        temperature: float
+          temperature in K at which the experimental measurement was carried out. By default: 298 K
         """
+        R = 1.9872042586408 * 0.001 * unit.kilocalorie_per_mole / unit.kelvin # Gas constant in kcal/mol/K
+        IC50_upper = IC50 * unit.nanomolar + uncertainty * unit.nanomolar
+        IC50_lower = IC50 * unit.nanomolar - uncertainty * unit.nanomolar
```
richardjgowers, line 80: we can tweak this now we've looked up the proper way.  maybe include the link to the solution we found so it's obvious what we're doing here

cinnabar/measurements.py
```diff
@@ -52,19 +54,52 @@ class Measurement(DefaultModel):
 
     @classmethod
     def from_experiment(cls,
-                        pIC50: float,
+                        IC50: float,
                         label: str,
+                        uncertainty: float = 0,
                         source: str = '',
+                        temperature: float = 298,
```
IAlibay, line 61: It's minor but I'd use 298.15 here instead, so we're consistent with everywhere else.

cinnabar/measurements.py
```diff
@@ -52,19 +54,52 @@ class Measurement(DefaultModel):
...
+          label for this data point.
+        uncertainty: float, optional
+          uncertainty of the experimental value in nM, default is zero if no uncertainty is provided
+        source: str, optional
+          source of experimental measurement
+        temperature: float
+          temperature in K at which the experimental measurement was carried out. By default: 298 K
         """
+        R = 1.9872042586408 * 0.001 * unit.kilocalorie_per_mole / unit.kelvin # Gas constant in kcal/mol/K
+        IC50_upper = IC50 * unit.nanomolar + uncertainty * unit.nanomolar
```
IAlibay, line 79: This is a different approach to how we do it in the PLB, I would suggest using that approach so we're consistent unless this one has some kind of advantage.

IAlibay (changes requested):
Couple of quick comments

cinnabar/measurements.py
hannahbaumann: @richardjgowers Looking at the code in the PLB, there are conversion between pKi, Ki, IC50 and DG. Would it make sense to use some of that code/extend the conversion here or do you think this is enough (once fixed)?
https://github.com/openforcefield/protein-ligand-benchmark/blob/a0b67b93c4f12f63a8bb417a1f6e4df4dc10c6c5/plbenchmark/utils.py#L85C1-L85C87

IAlibay:
Cycling PR to see if it updates from recent commit.

cinnabar/tests/test_measurements.py
```diff
@@ -13,6 +14,18 @@ def test_ground():
     assert g3 == g4
 
 
+def test_Ki_to_DG():
```
IAlibay, line 17: Thanks! Maybe one thing to also test here is to try out varying the temperature (that way we know it's working as intended).

If you look at the parameterize fixture (see: https://github.com/OpenFreeEnergy/openfe/blob/0e37fa53362d6e69a33960e00711a4eb35e8670c/openfe/tests/protocols/test_openmmutils.py#L25-L35) you can pass in a list of values which can then be picked up by your test.

hannahbaumann: Added that!

cinnabar/measurements.py
```diff
@@ -53,19 +55,51 @@ class Measurement(DefaultModel):
...
         """
+        R = 1.9872042586408 * 0.001 * unit.kilocalorie_per_mole / unit.kelvin # Gas constant in kcal/mol/K
+        # Convert Ki and uncertainty to molar, then strip the unit off for the math
+        Ki = Ki.to(unit.molar).m
+        uncertainty = uncertainty.to(unit.molar).m
+        if Ki > 0:
+            # dG = RT ln Ki
+            DG = R.m * temperature.m * math.log(Ki) * unit.kilocalorie_per_mole
+        else:
+            raise ValueError(
```
IAlibay, line 88: I think it's the final thing. Could you add two tests to check the ValueErrors being raised here and on line 96?

IAlibay (changes requested):

cinnabar/measurements.py
hannahbaumann: Added tests for the two ValueErrors.

.idea/.gitignore
IAlibay, line 1: oo these probably shouldn't be here, could you remove these and add .idea to the global gitignore?

examples/example_notebook_cinnabar.ipynb
IAlibay, line 1: This is new, should it be part of this PR?

cinnabar/tests/test_from_experiment.py
```diff
@@ -0,0 +1,14 @@
+from openff.units import unit
+import pytest
+from cinnabar import measurements
+
+def test_Ki_to_DG():
```
IAlibay, line 5: Isn't this already under `test_measurements`?

cinnabar/tests/test_measurements.py
```diff
@@ -12,6 +13,30 @@ def test_ground():
...
+def test_Ki_to_DG(Ki, uncertainty, dG, dG_uncertainty, label, temp):
+
+    Ki_to_DG = cinnabar.Measurement.from_experiment(Ki, label, uncertainty, '', temp)
+
+    assert pytest.approx(dG, 0.001) == Ki_to_DG.DG
+    assert pytest.approx(dG_uncertainty, 0.01) == Ki_to_DG.uncertainty
+
+def test_negative_Ki():
+
+    with pytest.raises(ValueError, match=r"Ki value cannot be zero or negative. Check if dG value was provided instead of Ki."):
```
IAlibay, line 33: We try (and often fail to), stick to the pep8 requirements (i.e. in this case 2 empty lines between functions and a maximum of 79 characters). Could you ammend these to match this?

IAlibay (changes requested):

examples/example_notebook_cinnabar.ipynb
hannahbaumann: No, removed this.

cinnabar/tests/test_from_experiment.py
hannahbaumann: Yes, that was a mistake, removed this.

cinnabar/tests/test_measurements.py
hannahbaumann: Changed this!

cinnabar/measurements.py
```diff
@@ -53,19 +54,55 @@ class Measurement(DefaultModel):
...
+            temperature in K at which the experimental measurement was carried out.
+            By default: 298 K (298.15 * unit.kelvin)
         """
+        R = 1.9872042586408 * 0.001 * unit.kilocalorie_per_mole / unit.kelvin
+        # Convert Ki and uncertainty to molar, then strip unit off for the math
+        Ki = Ki.to(unit.molar).m
+        uncertainty = uncertainty.to(unit.molar).m
+        if Ki > 0:
+            # dG = RT ln Ki
+            DG = R.m * temperature.m * math.log(Ki) * unit.kilocalorie_per_mole
```
IAlibay, line 87:
```suggestion
        if Ki > 0 * unit.molar:
            DG = (unit.molar_gas_constant * temperature.to(unit.kelvin)
                  * math.log( Ki / unit.molar)).to(unit.kilocalorie_per_mole)
```
I'm just checking against what the PLB returns and I noticed we've got a 2 d.p. discrepancy. I _think_ it's because we're using slightly different values for `R`. We should pin this to the openff.units value just so that we can propagate changes where needed.

cinnabar/measurements.py
```diff
@@ -53,19 +54,55 @@ class Measurement(DefaultModel):
...
+        else:
+            raise ValueError(
+                "Ki value cannot be zero or negative. "
+                "Check if dG value was provided instead of Ki."
+            )
+        # Convert Ki uncertainty into dG uncertainty: RT * uncertainty/Ki
+        # https://physics.stackexchange.com/questions/95254/the-error-of-the-natural-logarithm
+        if uncertainty >= 0:
+            uncertainty_DG = R.m * temperature.m * uncertainty / \
+                             Ki * unit.kilocalorie_per_mole
```
IAlibay, line 97:
```suggestion
        if uncertainty >= 0 * unit.molar:
            uncertainty_DG = (unit.molar_gas_constantt * temperature.to(unit.kelvin)
                              * uncertainty / Ki).to(unit.kilocalcorie_per_mole)
```
Please double check the output units, but I think that makes sense?

Also just a quick comment on line continuation whilst you're reading the PEP8 spec. Generally implicit line continuations (i.e. using brackets) are prefered over explicit ones (i.e. using `\`). Mostly because `/  \` tends to be really confusing 😅

cinnabar/tests/test_measurements.py
```diff
@@ -13,6 +14,38 @@ def test_ground():
...
+     0.059 * unit.kilocalorie_per_mole, 'lig', 298.15 * unit.kelvin],
+    [0.1 * unit.micromolar, 0.01 * unit.micromolar, -9.55 * unit.kilocalorie_per_mole,
+     0.059 * unit.kilocalorie_per_mole, 'lig', 298.15 * unit.kelvin],
+    [100 * unit.nanomolar, 10 * unit.nanomolar, -10.57 * unit.kilocalorie_per_mole,
+     0.066 * unit.kilocalorie_per_mole, 'lig', 330 * unit.kelvin],
+])
+def test_Ki_to_DG(Ki, uncertainty, dG, dG_uncertainty, label, temp):
+
+    Ki_to_DG = cinnabar.Measurement.from_experiment(Ki, label, uncertainty, '', temp)
+
```
IAlibay, line 28:
```suggestion

    assert Ki_to_DG.DG.units == unit.kilocalorie_per_mole
    assert Ki_to_DG.uncertainty.units == unit.kilocalorie_per_mole
```

A bit cautious, but because `pint` will do a conversion on equality, it's worth just double checking that the output units are what we expect them to be (particularly if we decide to jump to kj/mole later on, it'll make sure we have tests that check the unit return type).

cinnabar/measurements.py
hannahbaumann: Yes, that looks good!

cinnabar/measurements.py
hannahbaumann: Yes, that looks good! Should I commit the changes or are you doing that?

IAlibay:
Thanks! Sorry it's been such a long process 😅 hopefully it should get faster going forward.

PR #96, opened 2023-07-20
fix badge URL

### LINKED ISSUE #94: readthedocs is deeply unhappy
fix readthedocs

mikemhenry:
Fixes #94 

Yes I know it says arsenic, but it is the correct URL.


IAlibay (approved):
Is there no way to migrate the project?

mikemhenry:
I think it is like github where when you transfer things it just keeps the redirect urls, if it is an issue I can try and nuke it and re-make it

IAlibay:
> I think it is like github where when you transfer things it just keeps the redirect urls, if it is an issue I can try and nuke it and re-make it

I'm happy to live with this tbh.

PR #107, opened 2023-11-06
Add functions to calculate cycle closure

### LINKED ISSUE #86: feature: error per cycle graph
* traverse computational graph, find all cycles of size < n, for each calculate cycle closure error
* add nice function for plotting above data

hannahbaumann:
## Description

## Todos
  - [ ] Enable the comparison of multiple methods within the same FEMap object

## Status
- [x] Ready to go


cinnabar/femap.py
```diff
@@ -491,3 +491,70 @@ def draw_graph(self, title: str = "", filename: Union[str, None] = None):
...
+                # the sign of the DDG has to change
+                if (ligA, ligB) in list(edges):
+                    ddg = y[list(edges).index((ligA, ligB))]
+                elif (ligB, ligA) in edges:
+                    ddg = -y[list(edges).index((ligB, ligA))]
+                # sum up DDGs along cycle
+                sum_ddgs += ddg
+
+            # divide by sqrt of number of ligands in the cycle
+            # to get cycle closure error PER EDGE
```
jthorton: If this were per edge I would expect it to be over `len(cycle)`lets update the comment to something like normalise by sqrt(cycle length) to allow comparison across different length cycles?

jthorton: Could we also report the uncertainty-normalised closure as `abs(sum_ddgs) / sqrt(sum sigma_i^2)` where sigma_i is the uncertainty of the ith edge in the cycle, it might be useful to keep the uncertainty in mind when looking at these errors or is there any other standard way of doing this, something like a z-score for the error?

hannahbaumann:
cycle_closure

hannahbaumann:
pre-commit.ci autofix

cinnabar/femap.py
hannahbaumann: I added this!

cinnabar/femap.py
```diff
@@ -745,3 +746,117 @@ def draw_graph(self, title: str = "", filename: Union[str, None] = None):
             plt.show()
         else:
             plt.savefig(filename, bbox_inches="tight")
+
+    def get_cycle_closure(self, max_cycle_length: int = 5) -> list[tuple[str, float]]:
```
jthorton: How do you feel about calling this something like `get_cycle_closure_dataframe` so it aligns with the other methods on FEMap?

cinnabar/femap.py
```diff
@@ -745,3 +746,117 @@ def draw_graph(self, title: str = "", filename: Union[str, None] = None):
...
+
+        Returns
+        -------
+        pd.DataFrame
+            DataFrame with columns 'cycle', 'cc (kcal/mol)', sorted by
+            cycle closure error descending.
+        """
+        network = self.to_legacy_graph()
+        edge_ddg = {(a, b): d["calc_DDG"] for a, b, d in network.edges(data=True)}
+        edge_uncertainty = {(a, b): d["calc_dDDG"] for a, b, d in network.edges(data=True)}
```
jthorton: Can we use `get_relative_dataframe` as it handles multiple sources correctly and we want to drop `to_legacy_graph`.

cinnabar/femap.py
```diff
@@ -745,3 +746,117 @@ def draw_graph(self, title: str = "", filename: Union[str, None] = None):
...
+                # the sign of the DDG has to change
+                if (lig_a, lig_b) in edge_ddg:
+                    sum_ddgs += edge_ddg[(lig_a, lig_b)]
+                    sum_var += edge_uncertainty[(lig_a, lig_b)] ** 2
+                elif (lig_b, lig_a) in edge_ddg:
+                    sum_ddgs -= edge_ddg[(lig_b, lig_a)]
+                    sum_var += edge_uncertainty[(lig_b, lig_a)] ** 2
+                else:
+                    # Edge missing from network; skip this cycle
+                    break
```
jthorton: We also need to do this per source in the FEMap.

cinnabar/femap.py
```diff
@@ -745,3 +746,117 @@ def draw_graph(self, title: str = "", filename: Union[str, None] = None):
...
+        Parameters
+        ----------
+        max_cycle_length : int, optional
+            Only consider cycles up to this length. Default 5.
+
+        Returns
+        -------
+        pd.DataFrame
+            DataFrame with columns 'cycle', 'cc (kcal/mol)', sorted by
+            cycle closure error descending.
```
jthorton: Let's list the column names exactly as they appear, similar to the `get_relative_dataframe` function.

cinnabar/femap.py
```diff
@@ -745,3 +746,117 @@ def draw_graph(self, title: str = "", filename: Union[str, None] = None):
...
+
+        df = (
+            pd.DataFrame(rows, columns=["cycle", "cc (kcal/mol)", "cc_unc_normalized (kcal/mol)"])
+            .sort_values("cc (kcal/mol)", ascending=False)
+            .reset_index(drop=True)
+        )
+
+        return df
+
+    def get_cc_based_edge_statistics(self, max_cycle_length: int = 5) -> pd.DataFrame:
```
jthorton: Can we work this into the theme and get dataframe in the name so people know what to expect not sure how to shorten and add that though.

cinnabar/femap.py
```diff
@@ -745,3 +746,117 @@ def draw_graph(self, title: str = "", filename: Union[str, None] = None):
...
+        """
+        For each edge, report how many cycles it appears in and
+        the mean and max cycle closure error of those cycles.
+        Edges with high mean closure error across many cycles are
+        likely candidates for re-simulation.
```
jthorton: Make it clear that these are the simulated edges we are looking at, maybe we can add some interpretation of the results to the docs rather than the method? 
```suggestion
        For each simulated edge, report how many cycles it appears in and
        the mean and max cycle closure error of those cycles per source.
```

cinnabar/femap.py
```diff
@@ -745,3 +746,117 @@ def draw_graph(self, title: str = "", filename: Union[str, None] = None):
...
+
+        rows = []
+        for (a, b), ccs in edge_cycles.items():
+            rows.append(
+                {
+                    "ligandA": a,
+                    "ligandB": b,
+                    "n_cycles": len(ccs),
+                    "mean_cc (kcal/mol)": round(sum(ccs) / len(ccs), 3),
+                    "max_cc (kcal/mol)": round(max(ccs), 3),
```
jthorton: Add source to the output and sort by `["source", "mean_cc (kcal/mol)"]`

cinnabar/plotting.py
```diff
@@ -893,3 +894,45 @@ def ecdf_plot_all_DDGs(
...
+
+    if df.empty:
+        warnings.warn("No cycles found; skipping plot.")
+        return
+
+    errors = df["cc (kcal/mol)"]
+    n_cycles = len(errors)
+
+    fig, ax = plt.subplots(figsize=(5, 4))
+    ax.hist(errors, bins="auto", alpha=0.6, color="steelblue")
```
jthorton: Let the user pass in the bin width or the bins array? Or we should standardise on a width value.

cinnabar/plotting.py
```diff
@@ -893,3 +894,45 @@ def ecdf_plot_all_DDGs(
...
+    ----------
+    fe_map : FEMap
+        FEMap object containing the calculated edges.
+    filename : str, optional
+        If provided, the plot will be saved to this filename.
+    max_cycle_length : int, optional
+        Only consider cycles up to this length. Defaults to 5.
+        The matplotlib Figure object, which can be edited further.
+    """
+    df = fe_map.get_cycle_closure(max_cycle_length=max_cycle_length)
```
jthorton: We need to plot a histogram per source or let the user define a list of sources to plot.

cinnabar/femap.py
```diff
@@ -891,3 +892,115 @@ def draw_graph(self, title: str = "", filename: str | None = None):
             plt.show()
         else:
             plt.savefig(filename, bbox_inches="tight")
+
+    def get_cycle_closure(self, max_cycle_length: int = 5) -> list[tuple[str, float]]:
```
IAlibay:
```suggestion
    def get_cycle_closure(self, max_cycle_length: int = 5) -> pd.DataFrame:
```

cinnabar/femap.py
```diff
@@ -891,3 +892,115 @@ def draw_graph(self, title: str = "", filename: str | None = None):
...
+            else:
+                # Normalize by sqrt(cycle length) to allow comparison across
+                # different cycle lengths
+                cc = abs(sum_ddgs / math.sqrt(len(cycle)))
+                cc_uncertainty_normalized = abs(sum_ddgs) / math.sqrt(sum_var)
+                rows.append(
+                    {
+                        "cycle": tuple(cycle),
+                        "cc (kcal/mol)": round(cc, 2),
+                        "cc_unc_normalized (kcal/mol)": round(cc_uncertainty_normalized, 2),
```
IAlibay: Please correct me, but don't the units cancel out for this one?

cinnabar/femap.py
```diff
@@ -891,3 +892,115 @@ def draw_graph(self, title: str = "", filename: str | None = None):
...
+                    sum_var += edge_uncertainty[(lig_a, lig_b)] ** 2
+                elif (lig_b, lig_a) in edge_ddg:
+                    sum_ddgs -= edge_ddg[(lig_b, lig_a)]
+                    sum_var += edge_uncertainty[(lig_b, lig_a)] ** 2
+                else:
+                    # Edge missing from network; skip this cycle
+                    break
+
+            else:
+                # Normalize by sqrt(cycle length) to allow comparison across
```
IAlibay: Probably want to mention this in the docstring.

IAlibay: It may be useful to have the unnormalized variant on hand too?

IAlibay: Do we have some previously published information of someone using this? This roughly makes sense, but it's not immediately clear to me that it's standard practice.

cinnabar/femap.py
hannahbaumann: Changed this!

cinnabar/femap.py
hannahbaumann: Changed this!

cinnabar/femap.py
hannahbaumann: Changed this!

cinnabar/femap.py
hannahbaumann: Updated this!

cinnabar/femap.py
hannahbaumann: Changed this to `get_cycle_closure_edge_statistics_dataframe`

cinnabar/femap.py
hannahbaumann: Changed this!

cinnabar/plotting.py
hannahbaumann: Changed this, added a default parameter of 0.5 to the arguments.

cinnabar/plotting.py
hannahbaumann: Changed this. For now it's putting all histograms into a single plot, because I thought that maybe people don't have lots of sources, or should we do a separate plot per source?

hannahbaumann:
pre-commit.ci autofix

hannahbaumann:
pre-commit.ci autofix

cinnabar/plotting.py
```diff
@@ -1058,3 +1059,64 @@ def ecdf_plot_all_DDGs(
...
     )
+
+
+def plot_cycle_closure(
+    fe_map: FEMap,
+    filename: Optional[str] = None,
+    max_cycle_length: int = 5,
+    sources: Optional[list[str]] = None,
+    bin_width: float = 0.5,
+) -> plt.Figure:
```
hannahbaumann, line 1069:
```suggestion
) -> plt.Figure | None:
```

cinnabar/femap.py
hannahbaumann: That would be me in the SepTop paper =)
I think the idea (coming from Eric D.) was that the expected accumulated error scales at sqrt(n) if the errors are considered independent, giving an approximate contribution of a single edge to the cycle closure error.

hannahbaumann: Added this!

hannahbaumann: Added an explanation for the different cases to the doc string.

hannahbaumann:
pre-commit.ci autofix

cinnabar/femap.py
```diff
@@ -946,3 +947,157 @@ def draw_graph(
...
             fig.savefig(filename, bbox_inches="tight", dpi=300)
         plt.close(fig)
+
+    def get_cycle_closure_dataframe(self, max_cycle_length: int = 5) -> pd.DataFrame:
+        """
+        Calculate cycle closure errors for all cycles in the network.
+
+        Parameters
+        ----------
+        max_cycle_length : int, optional
```
jthorton:
```suggestion
        max_cycle_length : int, default 5
```

cinnabar/femap.py
```diff
@@ -946,3 +947,157 @@ def draw_graph(
...
+        )
+
+    def get_cycle_closure_edge_statistics_dataframe(self, max_cycle_length: int = 5) -> pd.DataFrame:
+        """
+        For each simulated edge, report how many cycles it appears in and
+        the mean and max cycle closure error of those cycles per source.
+
+        Parameters
+        ----------
+        max_cycle_length : int, optional
```
jthorton:
```suggestion
        max_cycle_length : int, default 5
```

cinnabar/plotting.py
```diff
@@ -1058,3 +1059,64 @@ def ecdf_plot_all_DDGs(
...
+    fe_map : FEMap
+        FEMap object containing the calculated edges.
+    filename : str, optional
+        If provided, the plot will be saved to this filename.
+    max_cycle_length : int, optional
+        Only consider cycles up to this length. Defaults to 5.
+        The matplotlib Figure object, which can be edited further.
+    sources : list[str], optional
+        List of sources to plot. If None, all sources are plotted.
+    bin_width : float, optional
```
jthorton:
```suggestion
    filename : str, default None
        If provided, the plot will be saved to this filename.
    max_cycle_length : int, default 5
        Only consider cycles up to this length. Defaults to 5.
        The matplotlib Figure object, which can be edited further.
    sources : list[str], default None
        List of sources to plot. If None, all sources are plotted.
    bin_width : float, default 0.5
```

cinnabar/plotting.py
```diff
@@ -1058,3 +1059,64 @@ def ecdf_plot_all_DDGs(
...
+
+    if df.empty:
+        warnings.warn("No cycles found; skipping plot.")
+        return None
+
+    if sources is not None:
+        df = df[df["source"].isin(sources)]
+        if df.empty:
+            warnings.warn(f"No cycles found for sources {sources}; skipping plot.")
+            return None
```
jthorton: the other types of plots all use `ValueError` when this type of validation fails, lets make it consistent and do the same here!

cinnabar/femap.py
```diff
@@ -946,3 +947,157 @@ def draw_graph(
...
+        -----
+        Three cycle closure metrics are calculated:
+
+        - ``cc (kcal/mol)``: the raw absolute sum of DDGs around the cycle. Units: kcal/mol.
+
+        - ``cc_per_edge (kcal/mol)``: the cycle closure divided by the square root of the cycle
+          length, to allow comparison across different cycle lengths. Units: kcal/mol.
+
+        - ``cc_unc_normalized``: the cycle closure error divided by its propagated uncertainty,
+          calculated as ``abs(sum_ddgs) / sqrt(sum_var)``.
```
jthorton, line 981: Do you want to add the septop paper as a reference in the doc string for the `cc_per_edge`?

cinnabar/tests/test_plotting.py
```diff
@@ -511,3 +512,50 @@ def test_plot_ecdf_colors(fe_map, tmp_path):
     # check that the line color matches the specified color
     line = fig.get_axes()[0].lines[0]
     assert line.get_color() == "#FF5733"
+
+
+def test_plot_cycle_closure(fe_map, tmp_path):
```
jthorton, line 516: Can we add a test for more than once source in the dataframes.

cinnabar/tests/test_femap.py
```diff
@@ -838,3 +839,48 @@ def test_missing_estimator_metadata(example_map):
...
+    """A perfect cycle with zero cycle closure."""
+    kcalpm = unit.kilocalorie_per_mole
+    fe = femap.FEMap()
+    fe.add_relative_calculation("A", "B", value=1.0 * kcalpm, uncertainty=0.1 * kcalpm)
+    fe.add_relative_calculation("B", "C", value=1.0 * kcalpm, uncertainty=0.1 * kcalpm)
+    fe.add_relative_calculation("C", "A", value=-2.0 * kcalpm, uncertainty=0.1 * kcalpm)
+    return fe
+
+
+def test_get_cycle_closure_perfect_cycle(perfect_cycle):
```
jthorton, line 844: Can we add a test for more than one source in the dataframe as well.

jthorton (changes requested):
Thanks @hannahbaumann, just a couple of minor things to add, and please add a news entry!

cinnabar/femap.py
```diff
@@ -946,3 +947,157 @@ def draw_graph(
...
+                    else:
+                        # Edge missing from network; skip this cycle
+                        break
+
+                else:
+                    # Normalize by sqrt(cycle length) to allow comparison across
+                    # different cycle lengths
+                    cc = abs(sum_ddgs)
+                    cc_per_edge = abs(sum_ddgs) / math.sqrt(len(cycle))
+                    cc_z_score = abs(sum_ddgs) / math.sqrt(sum_var)
```
IAlibay: If the uncertainty is 0 (e.g. if someone failed to feed it in), I believe this will cause a ZeroDivisionError.

cinnabar/femap.py
```diff
@@ -946,3 +947,157 @@ def draw_graph(
...
+        }
+
+        rows = []
+        for source, source_cc_df in cc_df.groupby("source"):
+            edge_ddg = edge_ddg_by_source.get(source, {})
+            edge_cycles: dict[tuple, list[float]] = defaultdict(list)
+
+            for _, row in source_cc_df.iterrows():
+                cycle = list(row["cycle"])
+                cc = row["cc_per_edge (kcal/mol)"]
```
IAlibay: Both the column names and docstring don't make it clear that you're using `cc_per_edge` (i.e. the cycle closure divided by sqrt(cycle_length).

It would be good to make that clearer.

cinnabar/femap.py
```diff
@@ -946,3 +947,157 @@ def draw_graph(
...
+        - n_cycles
+        - mean_cc (kcal/mol)
+        - max_cc (kcal/mol)
+
+        Sorted by source and mean cycle closure error descending.
+        """
+        from collections import defaultdict
+
+        cc_df = self.get_cycle_closure_dataframe(max_cycle_length=max_cycle_length)
+        comp_df = self.get_relative_dataframe()
```
IAlibay: You call `get_relative_dataframe` here but it's also called when you call `get_cyccle_closure_dataframe` is there any way to just directly use `cc_df`?

cinnabar/plotting.py
```diff
@@ -1058,3 +1059,64 @@ def ecdf_plot_all_DDGs(
         filename=filename,
         **kwargs,
     )
+
+
+def plot_cycle_closure(
+    fe_map: FEMap,
+    filename: Optional[str] = None,
```
IAlibay: Can you switch to str | None and avoid importing Optional & Union?

cinnabar/plotting.py
```diff
@@ -1058,3 +1059,64 @@ def ecdf_plot_all_DDGs(
...
+
+    Parameters
+    ----------
+    fe_map : FEMap
+        FEMap object containing the calculated edges.
+    filename : str, default None
+        If provided, the plot will be saved to this filename.
+    max_cycle_length : int, default 5
+        Only consider cycles up to this length. Defaults to 5.
+        The matplotlib Figure object, which can be edited further.
```
IAlibay: This seems like a typo?

cinnabar/plotting.py
```diff
@@ -1058,3 +1059,64 @@ def ecdf_plot_all_DDGs(
...
     )
+
+
+def plot_cycle_closure(
+    fe_map: FEMap,
+    filename: Optional[str] = None,
+    max_cycle_length: int = 5,
+    sources: Optional[list[str]] = None,
+    bin_width: float = 0.5,
+) -> plt.Figure | None:
```
IAlibay: Add a return entry to the docstring?

cinnabar/plotting.py
hannahbaumann: Changed this here and also above to raise a ValueError instead.

cinnabar/femap.py
hannahbaumann: added this!

cinnabar/tests/test_plotting.py
hannahbaumann: Added this!

cinnabar/tests/test_femap.py
hannahbaumann: Added this!

IAlibay (changes requested):
Couple of things

cinnabar/femap.py
hannahbaumann: Added a check for this!

cinnabar/femap.py
hannahbaumann: You're right, i think getting the edges directly from cc_df should also work, I updated this, unless I'm missing something? I also added a test that has multiple cycles to check for this.

hannahbaumann:
pre-commit.ci autofix

cinnabar/plotting.py
hannahbaumann: Ah, this must have come from not properly resolving a merge conflict, this was part of the return from the other function.. removed this here and added a proper returns below.

cinnabar/plotting.py
hannahbaumann: Done!

cinnabar/femap.py
hannahbaumann: Changed this!

cinnabar/plotting.py
hannahbaumann: Done!

cinnabar/femap.py
```diff
@@ -946,3 +947,154 @@ def draw_graph(
...
+        max_cycle_length : int, default 5
+            Only consider cycles up to this length. Default 5.
+
+        Returns
+        -------
+        The pandas DataFrame will have the following columns:
+        - source
+        - cycle
+        - cc (kcal/mol)
+        - cc_unc_normalized
```
IAlibay, line 967: Is this missing `cc_per_edge`?

cinnabar/plotting.py
```diff
@@ -1,4 +1,5 @@
 import itertools
+import warnings
```
IAlibay: Is this unused?

cinnabar/tests/test_plotting.py
```diff
@@ -1,4 +1,5 @@
 import matplotlib.pylab as plt
+import networkx as nx
```
IAlibay: Is this unused?

cinnabar/femap.py
```diff
@@ -946,3 +947,154 @@ def draw_graph(
...
+            edge_ddg = {(row["labelA"], row["labelB"]): row["DDG (kcal/mol)"] for _, row in source_df.iterrows()}
+            edge_uncertainty = {
+                (row["labelA"], row["labelB"]): row["uncertainty (kcal/mol)"] for _, row in source_df.iterrows()
+            }
+
+            network = nx.DiGraph()
+            for a, b in edge_ddg:
+                network.add_edge(a, b)
+
+            cycles = [c for c in nx.simple_cycles(network.to_undirected()) if len(c) <= max_cycle_length]
```
IAlibay, line 1000: I've been staring at this for a while and it's not immediately clear to me if calling `simple_cycles` on an undirected graph won't yield the same cycle in both orientation - i.e. [A, B, C] and [A, C, B].

Would this be intended behaviour? If not, should `cycle_basis` be used instead?

IAlibay (commented):
Couple of small things - mainly a question on simple_cyles and maybe if cycle_basis should be used instead?

cinnabar/femap.py
jthorton, line 967:
```suggestion
        - cc_per_edge (kcal/mol)
        - cc_unc_normalized
```

cinnabar/femap.py
jthorton: Wouldn't this show up in `test_get_cycle_closure_perfect_cycle`, there we check that the resulting dataframe has length 1 so only a single cycle is found.

hannahbaumann: It looks like `simple_cycles` filters those out:

> A “simple cycle”, or “elementary circuit”, is a closed path where no node appears twice. In a directed graph, two simple cycles are distinct if they are not cyclic permutations of each other. In an undirected graph, two simple cycles are distinct if they are not cyclic permutations of each other nor of the other’s reversal.
https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cycles.simple_cycles.html

But I could also switch to `cycle_basis` if that would be better overall!

cinnabar/tests/test_plotting.py
hannahbaumann: Removed!

cinnabar/plotting.py
hannahbaumann: Removed!

cinnabar/femap.py
hannahbaumann: Added this!

cinnabar/femap.py
jthorton: Thinking about this more, maybe `cycle_basis` would be better. Take the following example from the networkx docs:

```python
G = nx.Graph()
nx.add_cycle(G, [0, 1, 2, 3])
nx.add_cycle(G, [0, 3, 4, 5])
list(nx.simple_cycles(G))
> [[0, 1, 2, 3], [0, 1, 2, 3, 4, 5], [0, 3, 4, 5]]
```
Would we want the values on the smallest possible cycles of the graph? The super cycles formed by combinations of cycles don't tell us anything new do they?

hannahbaumann: I think my thought process from back when I was doing that as a grad student was that if we included more cycles (so also the super cycles) the risk of "missing" bad edges would be lower where by missing I mean that some cycles can give a low cycle closure by chance (bad edges but cancelation of errors). And then I included the `max_cycle_length` since the super large cycles don't seem very meaningful. In addition I had used this to then count how often an edge appears in a bad cycle, and there the super cycles are helpful to pin down the bad egg(s) in the cycle.
But maybe you're right and including those super cycles wouldn't really add that information?

jthorton: Okay that makes sense and seems to agree with the observations in https://pubs.acs.org/doi/10.1021/acs.jcim.5c00554 where closing the basis cycles does not mean the larger cycles will close, lets go with simple_cycles and come back to this if we find any issues!

jthorton (approved):
Thanks @hannahbaumann LGTM!

cinnabar/femap.py
IAlibay: It would be good to record this somewhere in an issue - otherwise we might never come back to it / remember why we made this decision.

IAlibay: Probably should document this cycle choosing behaviour too somewhere user facing (it'd be ok as a separate issue / PR).

IAlibay (approved):
I'll approve but I would like it if we could document some of the decision making on how we gather cycles.

cinnabar/femap.py
```diff
@@ -946,3 +947,155 @@ def draw_graph(
...
```
jthorton, line 1000: Wait do we want this to be `undirected`, this will miss self-loops if users run forward and backward edges which might be nice to include in the output?

cinnabar/femap.py
hannahbaumann: Opened an issue here: https://github.com/OpenFreeEnergy/cinnabar/issues/218

cinnabar/femap.py
hannahbaumann: As discussed offline, the directed graph would not catch something like this as a cycle: A->B; C->B; C->A.
For now we will leave it as is, but opening an issue that this should be fixed in the future (https://github.com/OpenFreeEnergy/cinnabar/issues/219)

jthorton: Spoke on slack about this, this does not pick up cycles in the directed case for cycles like `A->B; C->B; C->A?` we will update the notes of the function to say this does not cover self loops.

cinnabar/femap.py
```diff
@@ -946,3 +947,155 @@ def draw_graph(
...
+        Three cycle closure metrics are calculated:
+
+        - ``cc (kcal/mol)``: the raw absolute sum of DDGs around the cycle. Units: kcal/mol.
+
+        - ``cc_per_edge (kcal/mol)``: the cycle closure divided by the square root of the cycle
+          length, to allow comparison across different cycle lengths;
+          see Baumann et al. (DOI 10.1021/acs.jctc.3c00282). Units: kcal/mol.
+
+        - ``cc_unc_normalized``: the cycle closure error divided by its propagated uncertainty,
+          calculated as ``abs(sum_ddgs) / sqrt(sum_var)``.
```
hannahbaumann, line 981:
```suggestion
          calculated as ``abs(sum_ddgs) / sqrt(sum_var)``.
          
      The function currently does not consider self loop edges, e.g. A-->B and B-->A edges. 
```

cinnabar/femap.py
hannahbaumann, line 1000:
```suggestion
            # Using the undirected graph means that self loop edges are not considered. 
            cycles = [c for c in nx.simple_cycles(network.to_undirected()) if len(c) <= max_cycle_length]
```

cinnabar/femap.py
```diff
@@ -946,3 +947,158 @@ def draw_graph(
...
+          
+      The function currently does not consider self loop edges, e.g. A-->B and B-->A edges.
```
hannahbaumann:
```suggestion
        - ``cc_unc_normalized``: the cycle closure error divided by its propagated uncertainty,
          calculated as ``abs(sum_ddgs) / sqrt(sum_var)``.
          
        The function currently does not consider self loop edges, e.g. A-->B and B-->A edges. 
```

hannahbaumann:
pre-commit.ci autofix

PR #122, opened 2024-06-05
Add fraction best ligands metric

matthewtwarren:
## Description
Implementation of Chris Bayly's fraction of best ligands ranking/classification metric. 

This involves first computing the overlap coefficient - the fraction of N most-active experimental ligands that are in the N most-active predictions. This is implemented via the `_compute_overlap_coefficient` helper function, which relies on a 2D histogram constructed via the `_create_2d_histogram` function.

The fraction of best ligands is then calculated by averaging the overlap coefficients from N = 1 up to a given fraction of the ligands. The default is 0.5, i.e. up to num_ligands/2, rounded down. This is implemented in via `compute_fraction_best_ligands`. 

## Todos
  - [x] Implement helper functions to generate histogram and compute overlap coefficient
  - [x] Implement main function to compute metric
  - [ ] Write tests 

## Status 
- [x] Testing
- [ ] Ready to go

### DIFF
--- a/cinnabar/classification_metric.py
+++ b/cinnabar/classification_metric.py
@@ -5,6 +5,7 @@ from typing import Iterable
 from scipy import stats
 import numpy as np
 import ast
+import math
 
 
 def _experiment_prediction_binning(experiment_dG: Iterable[float], predict_dG: Iterable[float], n_classes:int =2, best_class_fraction:float=None):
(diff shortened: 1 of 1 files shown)


RiesBen:
HI @matthewtwarren , 
thanks for pushing this, I will have a look at it soon! :)

jthorton:
Thanks for the great work on this @matthewtwarren this is now completed in #164 in which we list you as an author!

PR #128, opened 2025-01-07
fix mambaforge CI error

atravitz:
## Description
fix failing CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
           channel-priority: true
           environment-file: devtools/conda-envs/env.yaml
           use-mamba: true
-          miniforge-variant: Mambaforge3
+          miniforge-variant: Miniforge3
 
       - name: Additional info about the build
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           channel-priority: true
           environment-file: devtools/conda-envs/env.yaml
           use-mamba: true
-          miniforge-variant: Mambaforge
+          miniforge-variant: Miniforge3
```
IAlibay, line 47: https://github.com/conda-incubator/setup-miniconda?tab=readme-ov-file#example-10-miniforge

Looks like docs say you should just set miniforge-version and it'll work.

IAlibay: (note this is what we do elsewhere, i.e. in MDA)

IAlibay (approved):

IAlibay:
rtd failure seems like a whole other issue since it's complaining about a missing path - probably worth dealing with elsewhere

PR #130, opened 2025-04-21
Test numpy 2.0

IAlibay:
## Description

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/devtools/conda-envs/env.yaml
+++ b/devtools/conda-envs/env.yaml
@@ -6,7 +6,7 @@ dependencies:
   - python
   - pip
   - networkx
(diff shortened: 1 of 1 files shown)


IAlibay:
I'm not sure we actually have to merge this, technically we should be able to just leave it as "numpy" in the recipe, but without forccing it conda-forge tries to resolve to numpy 1.26

mikemhenry:
it doesn't hurt to merge it in so we don't accidentally break things by adding a dep later that forces numpy 1.x

PR #158, opened 2025-08-29
remove old arsenic code

jthorton:
## Description
Remove references to arsenic in the code.

## Todos
  - [x]  rename arsenic to cli.py

## Questions
- [ ] Question1

## Status
- [x] Ready to go


jthorton:
This is a breaking change (not that I think others were importing arsenic directly) so it will probably need to be in the changelog, I'll add rever first then come back to this.

atravitz (approved):

PR #174, opened 2026-01-16
Add compare and ranking function for FEMaps

jthorton:
## Description
This PR adds a general compare and ranking function for a collection of FEMaps which contain results for the same edges which should be compared for any significant differences in performance. The FEMaps are compared by the user chosen metric, by default the mean unsigned error,  via the distribution of differences of that metric under a joint bootstrapping procedure. A two-sided p-value is then determined via the fraction of negative or positive differences in the metric (whichever is smaller), a multitest correction scheme is then used to correct the p-values which are used to rank the FEMaps using the compact letter display (CLD) assigned via the insert-absorb method. 

The result is two pandas dataframes, the first contains all evaluated metrics for each FEMap with confidence intervals and the CLD rank, the second contains the raw comparison data of the metric including p-values.

Example stats dataframe
```
    Model       MUE  MUE_CI_Lower  ...  KTAU_CI_Lower  KTAU_CI_Upper  CLD
0  FE Map 1  9.326389      9.055143  ...       0.407202       0.724801    a
1  FE Map 2  9.326389      9.062688  ...       0.377350       0.713957    a
2  FE Map 3  9.326389      8.987990  ...       0.044224       0.456656    b
```

Example comparsion
```
    Model 1   Model 2  Diff in rho  ...  p-value  significant  p-value corrected
0  FE Map 1  FE Map 2     0.002788  ...    0.666        False              0.666
1  FE Map 1  FE Map 3     0.321357  ...    0.002         True              0.006
2  FE Map 2  FE Map 3     0.318568  ...    0.002         True              0.006
```

## Todos
  - [x] add more tests
  - [ ] add a centralise/ shift method to the nodewise comparisons?

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready to go

Tips


cinnabar/compare.py
```diff
@@ -0,0 +1,275 @@
...
+
+import numpy as np
+import pandas as pd
+
+from cinnabar import FEMap
+from cinnabar.stats import AVAILABLE_STATS
+
+
+def compare_and_rank_femaps(
+    femaps: list[FEMap],
```
hannahbaumann: I wonder if it would be easier to have a single FEMap object that contains calc data from multiple experiments, using the `"source"` argument to distinguish between?

fjclark:
Commenting as Josh pointed me here.

I'm excited about this and would love to see this functionality in `cinnabar` soon! We've also been using the bootstrapping method from https://www.nature.com/articles/s42004-025-01428-y and it would be great to have it implemented here.

My main comment is I'm confused about when to use each of the `prediction_type` options and when they're valid. I think my confusion could be phrased as: "exactly which distribution are we trying to approximate with bootstrapping, and how much do we violate the IID assumption of bootstrapping for this distribution?". If we start by assuming each "single calculation" (dG for ABFE and ddG for RBFE) is an IID sample from the distribution we're interested in, then to me:

- For ABFE, `nodewise` makes sense (as used in the original paper). As I understand it,`edgewise` and `pairwise` would introduce correlation (each dG can contribute to many ddGs).
- For RBFE, `edgewise` makes sense for ddG stats. `nodewise` feels wrong because we're no longer resampling single calculations. Our dG estimates are fixed by the MLE from a fixed set of ddG estimates, so the bootstrapped distribution of dGs is generated by a different "process" to the `edgewise` ddG stats. An (impractical) approach which would be more intuitive to me would be to bootstrap over edges in a way which keeps a valid graph each time and recompute the dGs through the MLE each time to get the dG distribution. However, I guess this is the only practical way to use bootstrapping here. `pairwise` would introduce correlation because each ddG would contribute to many of the pairwise ddGs.

@jthorton Am I misunderstanding? What's the intended use of `pairwise`? Thanks!

Though please feel free to ignore -- I am likely overthinking this!

cinnabar/compare.py
jthorton: Thanks both, I have updated this now to use a single FEMap with different sources.

jthorton:
pre-commit.ci autofix

jthorton:
Thanks @fjclark this is great feedback. I agree that naively applying this to all of the available analysis types in Cinnabar is not a good idea and so I have removed the pairwise option and changed the default to edgewise, as I expect RBFE results to be the main use of the function. I agree that the nodewise analysis with the back-calculated absolute values is problematic due to the correlated nature of the values. For now, I recommend using the edgewise analysis for RBFE results in the docs and we can look into other possible methods in future to assess differences in ranking, hopefully this can serve as a good starting point though!

cinnabar/compare.py
```diff
@@ -0,0 +1,294 @@
...
+                "CI Upper": upper,
+                "p-value": p_value,
+                "significant": p_value < 0.05,
+            }
+        )
+    comparison_df = pd.DataFrame(comparison_data)
+
+    # if we have more than 2 models, apply multiple testing correction
+    if len(sources) > 2:
+        from statsmodels.stats.multitest import multipletests
```
IAlibay, line 208: Is this a new dependency?

cinnabar/compare.py
```diff
@@ -0,0 +1,294 @@
...
+
+    # set metrics to compute based on best practices if not provided
+    if metrics_to_compute is None:
+        if prediction_type == "edgewise":
+            metrics_to_compute = ["MUE", "RMSE"]
+        elif prediction_type == "nodewise":
+            metrics_to_compute = ["MUE", "RMSE", "RAE", "R2", "rho", "KTAU", "PI"]
+
+    # we must compute the rank metric however it is possible to miss it
+    if rank_metric not in metrics_to_compute:
```
IAlibay, line 131: Maybe not here, but we should think about how best to nudge users away from doing bad things - i.e. maybe we can add a warning that tells folks that it's a bad idea to rank using correlation metrics if you're doing edgewise comparisons.

cinnabar/compare.py
```diff
@@ -0,0 +1,294 @@
...
+        comparison_data.append(
+            {
+                "Model 1": source_i,
+                "Model 2": source_j,
+                f"Diff in {rank_metric}": summary_df[summary_df["Model"] == source_i][rank_metric].values[0]
+                - summary_df[summary_df["Model"] == source_j][rank_metric].values[0],
+                "CI Lower": lower,
+                "CI Upper": upper,
+                "p-value": p_value,
+                "significant": p_value < 0.05,
```
IAlibay: Do we want users to set this significance value?

cinnabar/compare.py
```diff
@@ -0,0 +1,294 @@
...
+        The number of bootstrap samples to use for estimating confidence intervals, by default 1000.
+    confidence_level : float, optional
+        The confidence level for the intervals, by default 0.95.
+
+    Note
+    ----
+    - The comparison method uses a joint bootstrapping procedure that generates a distribution of differences in the rank metric and checks for significant differences using a method inspired by. [1]_
+    - Each source must be evaluated on the same set of edges.
+    - Prediction types "nodewise" and "edgewise" correspond to DGs and edgewise DDGs respectively.
+    - When we have more than 2 models, we apply multiple testing correction to the pairwise comparisons using the ``Holm`` method by default.
```
IAlibay: Do you have a small explanation of why Holm over any of the other methods?

cinnabar/tests/test_compare.py
```diff
@@ -0,0 +1,134 @@
...
+    summary_df, comparison_df = compare_and_rank_results(
+        fe_map,
+        prediction_type="nodewise",
+        rank_metric="PI",
+    )
+    for metric in ["MUE", "RMSE", "R2", "rho", "PI", "KTAU", "RAE"]:
+        assert metric in summary_df.columns
+        for ci in ["Upper", "Lower"]:
+            assert f"{metric}_CI_{ci}" in summary_df.columns
+    summary_df.to_csv("summary_df.csv")
```
IAlibay: Should this be writing to disk unprotected by a temporary directory?

cinnabar/tests/test_compare.py
```diff
@@ -0,0 +1,134 @@
+import numpy as np
+import pytest
+from openff.units import unit
+
+from cinnabar import FEMap
+from cinnabar.compare import compare_and_rank_results
+
+
+def test_compare_and_rank_results(fe_map):
+    np.random.seed(42)
```
IAlibay, line 10: Should the next test also be seeded / check the outcomes?

IAlibay (changes requested):
Sorry that took me so long to review. Overall this looks good to me, but I've got a few questions.

cinnabar/compare.py
jthorton: Yeah I agree, hopefully the docs guide them in the right direction but a warning would help!

cinnabar/compare.py
jthorton: Ah yes good catch, it was already present I think but I'll list it in the env!

cinnabar/compare.py
jthorton: Good idea added an option using the standard name `alpha`.

cinnabar/compare.py
jthorton: Added something and a link to the wiki

cinnabar/tests/test_compare.py
jthorton: Good spot thats left over from checking the output, removed!

cinnabar/tests/test_compare.py
jthorton: Good point, added the seed and expanded the test.

jthorton:
pre-commit.ci autofix

jthorton:
pre-commit.ci autofix

cinnabar/compare.py
```diff
@@ -0,0 +1,304 @@
...
+    # set metrics to compute based on best practices if not provided
+    if metrics_to_compute is None:
+        if prediction_type == "edgewise":
+            metrics_to_compute = ["MUE", "RMSE"]
+        elif prediction_type == "nodewise":
+            metrics_to_compute = ["MUE", "RMSE", "RAE", "R2", "rho", "KTAU", "PI"]
+
+    # we must compute the rank metric however it is possible to miss it
+    if rank_metric not in metrics_to_compute:
+        metrics_to_compute.append(rank_metric)
```
IAlibay, line 134: [nit] This does an in-place modification of the input kwarg. It's probably safe, but always better to avoid in case someone is using a pre-defined variable somewhere outside of the method call.

cinnabar/compare.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # summarize pairwise comparisons with corrected p-values
+    comparison_data = []
+    for (source_i, source_j), diffs in pairwise_metrics.items():
+        lower = np.percentile(diffs, (1 - confidence_level) / 2 * 100)
+        upper = np.percentile(diffs, (1 + confidence_level) / 2 * 100)
+        # calculate the p-value as the fraction of bootstrap samples that cross zero
+        # use a 2-tailed test
+        # inspired by https://www.nature.com/articles/s42004-025-01428-y
+        p_value = 2 * min(np.mean(diffs < 0), np.mean(diffs > 0))
```
IAlibay, line 191: If I understand this correctly, there is a chance that the p value ends up being 0 when you have a clear winner. Is this something that would be misleading to users?

cinnabar/compare.py
```diff
@@ -0,0 +1,304 @@
...
+            node_label = row["label"]
+            if not row["computational"]:
+                source_label = "exp"
+            else:
+                source_label = f"{row['source']}_calc"
+            predictions_by_key[node_label][source_label] = row["DG (kcal/mol)"]
+
+    elif prediction_type == "edgewise":
+        rel_df = femap.get_relative_dataframe()
+        sources = rel_df[rel_df["computational"] == True]["source"].unique()
```
IAlibay: On line 68 you to `to_list()` too, whilst here you only do `unique()`. Should the two be aligned?

cinnabar/compare.py
```diff
@@ -0,0 +1,304 @@
...
+    ----------
+    .. [1] https://www.nature.com/articles/s42004-025-01428-y
+    """
+
+    # get the predictions and experimental values from the FEMaps and align them into a DataFrame
+    predictions_by_key: dict[tuple[str, str] | str, dict[str, float]] = defaultdict(dict)
+    if prediction_type == "nodewise":
+        abs_df = femap.get_absolute_dataframe()
+        # get the computational sources we want to compare
+        sources = abs_df[abs_df["computational"] == True]["source"].unique().tolist()
```
IAlibay: Can't you just do:


```suggestion
        sources = abs_df[abs_df["computational"]]["source"].unique().tolist()
```
Same thing for line 93.

cinnabar/compare.py
```diff
@@ -0,0 +1,304 @@
...
+                "CI Lower": lower,
+                "CI Upper": upper,
+                "p-value": p_value,
+                "significant": p_value < alpha,
+            }
+        )
+    comparison_df = pd.DataFrame(comparison_data)
+
+    # if we have more than 2 models, apply multiple testing correction
+    if len(sources) > 2:
```
IAlibay, line 207: [nit] If you don't have > 2, you don't have a `p-value corrected` column. Would it be better for users to always have the same columns and just return the same as the p-value in that case or would it be better to not have the column in that case?

cinnabar/compare.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def compare_and_rank_results(
+    femap: FEMap,
+    prediction_type: Literal["nodewise", "edgewise"] = "edgewise",
+    rank_metric: Literal["MUE", "RMSE", "RAE", "R2", "rho", "KTAU", "PI"] = "MUE",
+    metrics_to_compute: list[Literal["MUE", "RMSE", "RAE", "R2", "rho", "KTAU", "PI"]] | None = None,
+    num_bootstraps: int = 1_000,
+    confidence_level: float = 0.95,
+    alpha: float = 0.05,
```
IAlibay, line 19: [nit] Some bounds checks on some of these arguments could be useful to avoid users passing bad things and getting odd failures.

cinnabar/compare.py
```diff
@@ -0,0 +1,304 @@
...
+    # we must compute the rank metric however it is possible to miss it
+    if rank_metric not in metrics_to_compute:
+        metrics_to_compute.append(rank_metric)
+
+    # check we can compute all requested metrics
+    for metric in metrics_to_compute:
+        if metric not in _AVAILABLE_STATS:
+            raise ValueError(f"Metric {metric} is not available.")
+
+    metrics_by_source = dict((source, dict((metric, []) for metric in metrics_to_compute)) for source in sources)
```
IAlibay: [nit] this feels like it would be easierr to parse visually

```suggestion
    metrics_by_source = {source: {metric: [] for metric in metrics_to_compute} for source in sources}
```

IAlibay (changes requested):
Couple of small things, I'm going to put a "request changes" for some of the questions but it's not majorly blocking.

IAlibay:
Notebook looks great - I don't have any comments about it.

cinnabar/compare.py
jthorton: Yes good catch, changed to `tolist()`

cinnabar/compare.py
jthorton: Thats right, I did find mention of a method which adds 1 to the numerator and denominator to avoid exact zero answers but I am not confident that this is the best practice. One option is to push users to use the reported metric differences and CI around those to interpret the results when this happens, I could add a warning to the docs about it?

cinnabar/compare.py
jthorton: Having the output be stable would be great but I don't want people to think a correction was applied when it wasn't. I think this is something we could easily add in future though if people want it?

cinnabar/compare.py
jthorton: Fixed to use a copy.

cinnabar/compare.py
jthorton: Added checks and tests.

jthorton:
pre-commit.ci autofix

cinnabar/compare.py
IAlibay: Works for me.

cinnabar/compare.py
IAlibay: Yeah warning would be good!

IAlibay (approved):
Looking through the remaining things, it looks like it's just that warnings in the docs. I'lll apprrove early, thanks!

PR #177, opened 2026-01-20
Raise an error for zero uncertainties in the MLE solver

### LINKED ISSUE #97: Error of exactly zero will trigger SVD failure when generating absolute values
If you somehow have an error of zero in your input to cinnabar, and then call `generate_absolute_values`, you'll end up with an SVD failure in np.linalg.pinv (alongside a bunch of NaN's along the way). 

We should capture this edge case and ensure that folks don't feed in such values, or at least pad the number by a tiny value.

jthorton:
## Description
Fixes #97 by padding calculated errors of exactally zero by some small amount.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready to go

Tips


jthorton:
pre-commit.ci autofix

IAlibay (commented):
Not 100% sure about the impact, but adding this padding *should* be ok?

cinnabar/stats.py
```diff
@@ -205,6 +205,13 @@ def mle(graph: nx.DiGraph, factor: str = "f_ij", node_factor: Union[str, None] =
             action="symmetrize",
             node_label=node_factor.replace("_", "_d"),
         )
+    # if the uncertainty is zero, the MLE solver will fail
+    # set all values to a small value if exactly zero
+    # see <https://github.com/OpenFreeEnergy/cinnabar/issues/97> for details
+    for i in range(N):
```
IAlibay: Should we be warning users that this is happening / we're fixing their data on the fly? In many ways the 1e-6 is likely going to come from cases where only a single repeat was used, so maybe we should warn folks that this isn't a great situation (given that MLE does use the errors).

IAlibay (changes requested):
Sorry that was more of a lightweight request changes.

cinnabar/stats.py
hannahbaumann: I'm also not super sure. With the way how we report the standard dev in the protocol results, I suspect that a lot of people will have zero uncertainties as they may only run one repeat in production. So I would say either a warning or if we raise an error we need very good documentation for how users get the MBAR errors into their "cinnabar plotting".

IAlibay (approved):
This makes sense, thanks!

What do you think about following up with some docs or cookbook in openfe that shows how to use the MBAR error instead when standard deviations aren't available?

PR #178, opened 2026-01-20
Raise an error when we have repeated edge results

### LINKED ISSUE #133: MLE DG estimates are wrong for bidirectional RBFE calculations on ligand network.
**Describe the bug**
I have run bidirectional RBFE calculations (lig1->lig2 and lig2->lig1 for all the transformations in the ligand network) on a dataset with a user-defined ligand network and when I compute and plot the predicted vs experimental dG values, I see that the predicted dG values are not calculated properly and I get very high values of MUE and RMSE (correct R2 though). This does not happen when I only use unidirectional calculations, in which case the pred. vs exp dG plot is very reasonable. I was wondering if this is known behaviour because the MLE algorithm is not designed to work on bidirectional ddG values? 

**To Reproduce**
I attach the cinnaber_input.csv file that I used.

[cinnabar_input.csv](https://github.com/user-attachments/files/20672180/cinnabar_input.csv)

I also attach the ligand network along with the predicted ddG and dG values.

Image

**Software versions**
  - My OS is Ubuntu 24.04
  - For FEP calculations and analysis I used OpenFE v1.3  

**Output**
The pred. vs exp. dG plot shows very high MUE and RMSE values and the ddG values do not add up to the estimated dGs (see ligand network image).

Image

**Expected behavior**
If I only use unidirectional ddG values in my cinnaber-input.csv the pred. vs exp. dG plot and the statistics look much more reasonable and in line with what we expected. 

Image

**Additional context**
I also tried to exclude from the calculation the edges where the forward and reverse ddG values did not match and I still did not get a correct dG prediction.

### LINKED ISSUE #136: Deviation in edge repeats not accounted for in computed uncertainties
Hello, see this example input with CAT-13n,CAT-13a edge repeat:

```
# Experimental block
# Ligand, expt_DDG, expt_dDDG
CAT-13a, -8.83 , 0.10
CAT-13n, -8.53 , 0.10

# Calculated block
# Ligand1,Ligand2, calc_DDG, calc_dDDG(MBAR), calc_dDDG(additional)
CAT-13n,CAT-13a,   -1.73	,0.12,0.0
CAT-13n,CAT-13a,   -5.00        ,0.12,0.0
```

Computed uncertainties with `generate_absolute_values` are not affected by the deviation in the ddGs between the repeats. My work around would be to do the averaging and stddev calculation prior to creating the input csv for cinnabar and stick the stddev between repeats into the cal_dDDG(additional) column

jthorton:
## Description
Fixes #133 & #136 by raising an error if we find repeated or bidirectional results in the input graph when trying to compute the absolute values with the MLE method. This now points the users to the docs and advises them to combine the results in some way rather than doing this automatically. 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Checklist

## Status
- [ ] Ready to go

Tips

### DIFF
--- a/cinnabar/femap.py
+++ b/cinnabar/femap.py
@@ -504,6 +504,8 @@ class FEMap:
         """
         # reduces to nx.DiGraph
         g = nx.DiGraph()
+        # the MLE method can only use a single result per edge, we need to raise and error if we have repeats or bidirectional results
+        edges_seen = []
         # add DDG values from computational graph
         for a, b, d in self._graph.edges(data=True):
             if not d["computational"]:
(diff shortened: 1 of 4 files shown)


IAlibay (approved):
This seems like a good patch to me, thanks!

Issue #147, opened 2025-08-26
Add pydantic-v2 support


jthorton:

jthorton:
The plan is use models in gufe to make the migration easier, however we are waiting for openmm to be made an optional dependecy of gufe to keep cinnabar lightweight!

jthorton:
We have now removed the use of pydantic in #190 as an alternative.

PR #190, opened 2026-02-20
Drop pydantic use

jthorton:
## Description
As an alternative to fixing #147, we drop Pydantic support and use a simple dataclass. Thanks, @atravitz, for the idea! The class is still frozen and we convert to the default units in the init to get the same behaviour 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready to go

Tips


cinnabar/tests/test_measurements.py
```diff
@@ -135,3 +135,16 @@ def test_is_true_ground(label, expected):
...
+
+def test_convert_default_units():
+    m = cinnabar.Measurement(
+        labelA="foo",
+        labelB="bar",
+        DG=4.0 * unit.kilojoule_per_mole,
+        uncertainty=1.0 * unit.kilojoule_per_mole,
+        computational=True,
+    )
+
```
atravitz:
```suggestion
 pytest.approx(0.9560, 0.0001)==m.DG.m
```
as-is, this test would pass if somehow just the units were updated, but the magnitude was not converted. Making it a bit more strict for my peace of mind.

cinnabar/measurements.py
```diff
@@ -85,23 +84,41 @@ class Measurement(DefaultModel):
...
+        temperature: unit.Quantity = 298.15 * unit.kelvin,
+    ):
+        """
+        Initialize a Measurement object converting all quantities to the correct default units.
+        """
+        object.__setattr__(self, "labelA", labelA)
+        object.__setattr__(self, "labelB", labelB)
+        object.__setattr__(self, "DG", DG.to(unit.kilocalorie_per_mole))
+        object.__setattr__(self, "uncertainty", uncertainty.to(unit.kilocalorie_per_mole))
+        object.__setattr__(self, "computational", computational)
```
atravitz: I think this is the right way to do this, although it takes away one of the benefits of using a dataclass (avoiding boilerplate code). But we still get the frozen dataclass features!

Another option is to do this the gufe way and use a regular class and initialize private attributes, then expose them as properties to preserve pseudo-immutability. 

e.g., for `DG`:

```
def __init__(self, ..., DG):
    self._DG = DG.to(unit.kilocalorie_per_mole)

@property
def DG(self):
    return self._DG

```


pinging @ianmkenney and @mikemhenry in case they have any clever insights or opinions.

At the end of the day, I think that a user is less likely to use `object.__setattr__(measurement, "DG", wrong_value)` than `measurement._DG = wrong_value`, so Josh's implementation is better.

atravitz (changes requested):
One comment and one suggestion!

cinnabar/measurements.py
jthorton: Yeah I wasn't sure what to do and followed the python docs on dataclass so happy to go with whatever others feel is best!

jthorton:
pre-commit.ci autofix

atravitz (changes requested):
just two nitpicks, one docs and one non-blocking. thanks!

cinnabar/measurements.py
```diff
@@ -85,23 +85,53 @@ class Measurement(DefaultModel):
...
+        # This dataclass used to be based on pydantic and could automatically convert units from strings
+        # we now do this manually to avoid the dependency and not break old behavior.
+        unit_values = [DG, uncertainty, temperature]
+        for i in range(len(unit_values)):
+            if isinstance(unit_values[i], str):
+                # convert inplace to a quantity with units
+                unit_values[i] = unit.Quantity(unit_values[i])
+
+            elif isinstance(unit_values[i], (float, int)):
+                raise ValueError("DG, uncertainty, and temperature values must have units. Check input.")
```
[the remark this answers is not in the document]
jthorton: Done!

atravitz (approved):

Issue #109, opened 2023-11-06
base class for MLE estimation-like methods


richardjgowers:
add base class for MLE-like operations that consumes the FEMap

then port existing MLE solution to this new base class

jthorton:
Completed in #193

PR #193, opened 2026-02-27
Add support for other estimators 

### LINKED ISSUE #189: Make estimators work with FEMaps with multiple sources
**Is your feature request related to a problem? Please describe.**
When comparing a few methods (say the same edges calculated with different force fields) it would be easier to do the anaylsis using a single FEMap with different sources as you only have to add the experimental reference data once. I think the estimator code (`generate_absolute_values`) needs to be updated to support multiple subnetworks with different sources. 

This would also allow you to do plots for aggregated results, for example, creating a single FEMap for all JACS systems and then plotting the all-to-all pairwise DDGs values. 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

jthorton:
## Description
Impliments #109 and fixes #189.

This PR makes the estimator used in `generate_absolute_values` customisable, now an estimator which is a subclass of the new abstract base class `Estimator` can be passed as an argument and applied to the graph. The estimators now also handle sub-networks of results differentiated by their `source` tag automatically. Each estimator can also store metadata on the FEMap such as the covariance matrix in the case of the MLE estimator for downstream analysis like that in #149. 

This *should* not break existing workflows as calls to `femap.generate_absolute_values()` with a single source will still run the MLE method and add measurements to the graph tagged as `MLE`. 

However, users can now specify the estimator using
```python
from cinnabar import estimators
mle = estimators.MLEEstimator()
femap.generate_absolute_values(mle)
```

Users can also generate estimates for more than a single source at once, though each sub-network must be weakly connected. In cases with more than one source, the calculated DG measurement source is replaced with a new tag of the form `{estimator}({source})`, so if the input source is `openff-sage` for one set of calculations the new source tag will be `MLE(openff-sage)`.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready to go

Tips


IAlibay (approved):
This looks good to me. It's hard to know how well this will work for other methods until we implement other methods, but I can see it working.

jthorton:
pre-commit.ci autofix

PR #194, opened 2026-03-02
Add predictive index metric

jthorton:
## Description
This PR adds the predictive index metric introduced by [Pearlman et al.](https://pubs.acs.org/doi/abs/10.1021/jm0100279) which could be another useful metric to report on DG predictions. 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready to go

Tips

### DIFF
--- a/cinnabar/stats.py
+++ b/cinnabar/stats.py
@@ -235,7 +235,7 @@ def calculate_predictive_index(y_true: np.ndarray, y_pred: np.ndarray) -> float:
 
         C_{ij} = \begin{cases}
             1 & \text{if } (E_{j} - E_{i})/(P_{j} - P_{i}) > 0 \\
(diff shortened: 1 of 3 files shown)


IAlibay (changes requested):
Couple of small things, otherwise it looks good.

It's not immediately clear to me that this metric has been thoroughly tested enough, but it doesn't matter too much on inclusion at this point.

One thought I've not been able to work out is whether or not this prediction is significantly if you use pKi / pIC50, i.e. does having smaller numbers lead to more useful indications of prediction performance? Or better yet, is there a reason the original authors decided on pKi/pIC50 rather than Ki/IC50.

IAlibay:
I see, thanks! I think using the estimated dG might get around some of these issues but we should maybe just keep it in mind when we get around to using this metric in practice.

IAlibay (approved):
lgtm, thanks!

PR #197, opened 2026-03-27
test python 3.13 in CI

atravitz:
## Description
openfe supports python 3.13 and uses cinnabar, so we should be testing 3.13 here as well.

We should also update the CI matrix to reflect openfe's - we don't need to test every version against every OS.

Tips

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,12 +27,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu', 'macos']
+        os: ['ubuntu-latest']
         python-version:
-          - "3.10"
           - "3.11"
           - "3.12"
           - "3.13"
+        include:
+          - os: "macos-latest"
+            python-version: "3.12"
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
+          - "3.13"
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
+          - "3.13"
```
jthorton, line 34: Thanks, should we be dropping 3.10 as well to match openfe?

jthorton:
Thanks for this @atravitz, do you plan on updating the os testing here or should we do this in another PR?

atravitz:
@jthorton - I went ahead and dropped python 3.10 from testing, and now only test macos on python 3.12 (3.12 is what all of our example notebooks etc. work on, so I consider it the most critical)

.github/workflows/ci.yml
```diff
@@ -27,11 +27,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu', 'macos']
+        os: ['ubuntu-latest']
```
jthorton:
```suggestion
        os: [ubuntu-latest]
```
do we need quotes for these strings?

.github/workflows/ci.yml
```diff
@@ -27,11 +27,14 @@ jobs:
...
         python-version:
-          - "3.10"
           - "3.11"
           - "3.12"
+          - "3.13"
+        include:
+          - os: "macos-latest"
```
jthorton, line 36:
```suggestion
          - os: macos-latest
```

jthorton:
@atravitz thanks but the CI doesn't seem to be starting not sure if this is a github issue or not?

atravitz:
> @atravitz thanks but the CI doesn't seem to be starting not sure if this is a github issue or not?

fixed!

.github/workflows/ci.yml
atravitz: I think either way works, but it's my personal preference to keep strings in quotes for consistency

jthorton (approved):
LGTM!

PR #201, opened 2026-04-09
Add CI to ecdf plots using bootstrapping 

jthorton:
## Description
This PR adds a CI to ecdf plots using bootstrapping. At each iteration, the errors are subsampled with replacement and a new ecdf is calculated. The lower and higher percentile values at each true data point are then calculated based on the given CI, which is then plotted as a filled area in the same colour as the sample data. 

An example is shown using the test data labelled `test1` and a variation with added noise labelled `test2`. 

image

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready to go

Tips


jthorton:
pre-commit.ci autofix

cinnabar/plotting.py
```diff
@@ -644,15 +650,36 @@ def ecdf_plot(
...
         sns.ecdfplot(data, label=label, **default_kwargs)
 
+        # estimate an error bounds via bootsrapping over the data
+        boot_ecdfs = []
+        # we first need to sort the data so its in the same order used in the ecdf plot
+        data = np.sort(data)
+        for _ in range(nbootstraps):
+            sample = np.random.choice(data, size=len(data), replace=True)
+            # calculate the ECDF for this sample at each point in the true sample data
+            boot_ecdf = [np.mean(sample <= x) for x in data]
```
hannahbaumann: Do you have approximate timing data on how much this slows things down?

cinnabar/plotting.py
```diff
@@ -595,6 +595,8 @@ def ecdf_plot(
     colors: list[str] | None = None,
     ecdf_kwargs: dict[str, Any] | None = None,
     filename: str | None = None,
+    nbootstraps: int = 1_000,
+    ci: float = 0.95,
```
hannahbaumann, line 599: Potentially add a check that it's not above 1, in case people think in percent.

hannahbaumann (commented):
Thanks @jthorton , this looks great! I was just wondering about the speed of the bootstrapping, did you notice whether this makes the plotting very slow?

IAlibay (commented):
The code looks reasonable to me, but my questions would mostly be on best practices - is this widely used / are there papers we can cite here?

cinnabar/plotting.py
jthorton: For the jacs dataset of ~300 edges it took about 4 seconds up from 1 so its a little slower.

cinnabar/plotting.py
jthorton: Done!

jthorton:
> is this widely used / are there papers we can cite here?

Not that I have seen. I have made it so that you can remove the estimate by setting the bootstrap iterations to 0 so users can opt out, though and I can add more details of the procedure in the docs which are still to come.

cinnabar/plotting.py
```diff
@@ -644,15 +654,37 @@ def ecdf_plot(
...
+        color = colors[i]
+        default_kwargs["color"] = color
 
         sns.ecdfplot(data, label=label, **default_kwargs)
 
+        # estimate an error bounds via bootsrapping over the data if the number of bootstraps is > 0
+        if nbootstraps > 0:
+            boot_ecdfs = []
+            # we first need to sort the data so its in the same order used in the ecdf plot
+            data = np.sort(data)
```
IAlibay: The comment here is throwing me off a little bit, if you want to keepp the same order as the ecdf plot, would you not want to sort before plotting?

cinnabar/plotting.py
```diff
@@ -644,15 +654,37 @@ def ecdf_plot(
...
+            for _ in range(nbootstraps):
+                sample = np.random.choice(data, size=len(data), replace=True)
+                # calculate the ECDF for this sample at each point in the true sample data
+                boot_ecdf = [np.mean(sample <= x) for x in data]
```
IAlibay: I'm being told (thanks claude) that it would be slightly more efficient to do:

```suggestion
                boot_ecdf = np.searchsorted(np.sort(sample), data, side="right") / len(sample)
```

Locally on a tiny tiny dataset I'm getting an 8x speedup (microseconds to microseconds), but I'm sure that can add up on larger datasets.

IAlibay (approved):
Two small things, I'll approve eagerly because the code is fine even without the changes.

cinnabar/plotting.py
jthorton: Yeah we can do,  it is done automatically by seaborn, but now it's needed for the CI estimation so let's just do it early!

cinnabar/plotting.py
jthorton: Very cool, once you get how searchsorted works this makes sense added a longer comment so I remember when we come back to this!

PR #206, opened 2026-05-20
Add ECDF plotting tutorial 

jthorton:
Fixes #180 by adding a tutorial for the ECDF plots, docs drafted by me and cleaned up by Claude!

This also changes the API for the ECDF plotting functions to work with a single FEMap. While writing the docs I discovered a bug in the relative dataframe function which would return the wrong experimental values if absolute values had been generated on the FEMap first. This is also fixed. 

## Description
Provide a brief description of the PR's purpose here.

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] Rework the ECDF plotting API to work with a single FEMap with more than one source
  - [x] Rework the ECDF plotting functions to work with the internal FEMap dataframes, requires #149 to finish the pairwise function
  - [x] Fix the relative dataframe bug, which gives the wrong values if absolute values have been generated first using the `generate_absolute_values()` function of the `FEMap`
  - [ ] 

## Questions
- [ ] Question1

## Checklist
- [x] Added a ``news`` entry for new features, bug fixes, or other user facing changes.

## Status
- [x] Ready to go

Tips
* Comment "pre-commit.ci autofix" to have pre-commit.ci atomically format your PR.
  Since this will create a commit, it is best to make this comment when you are finished with your work.

PR #208, opened 2026-05-22
Support pIC50 units on dataframes

jthorton:
This is work towards custom units on plots, we now allow users to select the dataframes to be in DG (kcal/mol) or pIC50 a common unit used in drug discovery. 

After this we plan on reworking all plotting functions to use the dataframes internally, which should allow users to easily swap the reported units, see #206 as an example for the ecdf plots.

This PR also includes the relative dataframe fix from #206

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready to go

Tips

### DIFF
--- a/cinnabar/conversion.py
+++ b/cinnabar/conversion.py
@@ -82,7 +82,8 @@ def convert_observable(
     if final_type not in valid_types:
         raise ValueError(f"Unknown final_type: {final_type}. Must be one of: {', '.join(valid_types)}")
     # validate that uncertainty is non-negative if provided
-    if uncertainty is not None and uncertainty < 0:
+    # use np.any so the check works for both scalar and array Quantity inputs
+    if uncertainty is not None and np.any(uncertainty.m < 0):
         raise ValueError("Uncertainty must be positive")
 
     # store the conversion functions by (original_type, final_type) in a dictionary for easy lookup
--- a/cinnabar/femap.py
+++ b/cinnabar/femap.py
@@ -65,19 +65,14 @@ def _convert_dg_df_to_pic50(
     """
     missing = [c for c in (value_col, uncertainty_col) if c not in df.columns]
     if missing:
-        raise ValueError(
-            f"Column(s) {missing} not found in dataframe. "
-            f"Available columns: {list(df.columns)}"
-        )
+        raise ValueError(f"Column(s) {missing} not found in dataframe. Available columns: {list(df.columns)}")
     # default units should always be kcal/mol in the FEMap
     values = df[value_col].to_numpy() * _kcalpm
     uncertainties = df[uncertainty_col].to_numpy() * _kcalpm
     converted_values, converted_uncertainties = convert_observable(values, "dg", "pic50", uncertainties, temperature)
     # make sure we do not change the column order the dataframe should feel the same
     col_order = [
-        new_value_col if c == value_col
-        else (new_uncertainty_col if c == uncertainty_col else c)
-        for c in df.columns
+        new_value_col if c == value_col else (new_uncertainty_col if c == uncertainty_col else c) for c in df.columns
     ]
     return df.drop(columns=[value_col, uncertainty_col]).assign(
         **{new_value_col: converted_values.m, new_uncertainty_col: converted_uncertainties.m}
--- a/cinnabar/femap.py
+++ b/cinnabar/femap.py
@@ -552,6 +552,7 @@ class FEMap:
         - ``uncertainty (kcal/mol)`` / ``uncertainty (ΔpIC50)``
         - source
         - computational
+
         The dataframe will be sorted by source, computational, labelA, and labelB to ensure that pairing order is consistent.
         If `symmetrical` is True, the dataframe will include both (labelA, labelB) and (labelB, labelA) for each pair of labels, with opposite signs for DDG and the same uncertainty.
         If an estimator is used to generate the absolute binding affinities from relative results this function attempts
--- a/cinnabar/femap.py
+++ b/cinnabar/femap.py
@@ -63,13 +63,21 @@ def _convert_dg_df_to_pic50(
         Copy of the dataframe with ``value_col`` / ``uncertainty_col`` dropped and the two
         new columns in their place.
     """
+    missing = [c for c in (value_col, uncertainty_col) if c not in df.columns]
+    if missing:
+        raise ValueError(
+            f"Column(s) {missing} not found in dataframe. "
+            f"Available columns: {list(df.columns)}"
+        )
     # default units should always be kcal/mol in the FEMap
     values = df[value_col].to_numpy() * _kcalpm
     uncertainties = df[uncertainty_col].to_numpy() * _kcalpm
     converted_values, converted_uncertainties = convert_observable(values, "dg", "pic50", uncertainties, temperature)
     # make sure we do not change the column order the dataframe should feel the same
     col_order = [
-        new_value_col if c == value_col else (new_uncertainty_col if c == uncertainty_col else c) for c in df.columns
+        new_value_col if c == value_col
+        else (new_uncertainty_col if c == uncertainty_col else c)
+        for c in df.columns
     ]
     return df.drop(columns=[value_col, uncertainty_col]).assign(
         **{new_value_col: converted_values.m, new_uncertainty_col: converted_uncertainties.m}
@@ -401,9 +409,9 @@ class FEMap:
 
         Parameters
         ----------
-        observable_type : ANALYSIS_UNITS, optional
+        observable_type : {"dg", "pic50"}, default "dg"
             The observable type to report values in.  Defaults to ``dg`` (kcal/mol).
-            Use ``pic50`` to report ΔpIC50 values.
+            Use ``pic50`` to report DpIC50 values.
         temperature : Quantity, optional
             Temperature used for the unit conversion.  Defaults to 298.15 K.
 
@@ -411,12 +419,12 @@ class FEMap:
         ----
         The pandas DataFrame will have the following columns:
 
-        - labelA
-        - labelB
-        - ``DDG (kcal/mol)`` / ``ΔpIC50`` — depending on ``observable_type``
-        - ``uncertainty (kcal/mol)`` / ``uncertainty (ΔpIC50)``
-        - source
-        - computational
+        - ``labelA``
+        - ``labelB``
+        - ``DDG (kcal/mol)`` / ``DpIC50`` — depending on ``observable_type``
+        - ``uncertainty (kcal/mol)`` / ``uncertainty (unitless)``
+        - ``source``
+        - ``computational``
 
         Only simulated relative results are included for the computational results.
         The dataframe is sorted by source, computational, labelA, and labelB to ensure consistent ordering of results between sources.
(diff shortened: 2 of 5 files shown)


cinnabar/conversion.py
```diff
@@ -82,7 +82,8 @@ def convert_observable(
     if final_type not in valid_types:
         raise ValueError(f"Unknown final_type: {final_type}. Must be one of: {', '.join(valid_types)}")
     # validate that uncertainty is non-negative if provided
-    if uncertainty is not None and uncertainty < 0:
+    # use np.any so the check works for both scalar and array Quantity inputs
+    if uncertainty is not None and np.any(uncertainty.m < 0):
```
jthorton, line 86: This makes the unit conversion miuch easier on the dataframe as we can pass in a numpy array to the functions.

cinnabar/femap.py
```diff
@@ -346,20 +397,34 @@ def add_absolute_calculation(
...
+
         Note
         ----
         The pandas DataFrame will have the following columns:
+
         - labelA
         - labelB
-        - DDG (kcal/mol)
-        - uncertainty (kcal/mol)
+        - ``DDG (kcal/mol)`` / ``ΔpIC50`` — depending on ``observable_type``
```
jthorton: Should this be changed from `ΔpIC50` to `DpIC50` to make it easier to pull the values out rather than needing to find and copy and paste Δ each time?

jthorton:
pre-commit.ci autofix

cinnabar/femap.py
```diff
@@ -346,20 +392,34 @@ def add_absolute_calculation(
         )
         self.add_measurement(m)
 
-    def get_relative_dataframe(self) -> pd.DataFrame:
+    def get_relative_dataframe(
+        self,
+        observable_type: ANALYSIS_UNITS = "dg",
```
jthorton, line 425: Do we want to add a check that the provided `observable_type` is one of `dg` or `pic50` ?

cinnabar/tests/test_conversion.py
```diff
@@ -156,6 +156,10 @@ def test_convert_value_from_ic50(value, uncertainty, temp, output_type, expected
     ],
 )
 def test_convert_value_from_pic50(value, uncertainty, temp, output_type, expected_value, expected_uncertainty):
+    d = unit("dimensionless")
```
IAlibay, line 159: Why is this necessary? - see comment below on the behaviour change.

cinnabar/femap.py
IAlibay: Yes, right now the check will just silently fail if you pass in any other value (on line 465) - at least some kind of `else` would be good.

cinnabar/femap.py
IAlibay: We are using DDG not ΔΔG, so we should be consistent and call it DpIC50.

cinnabar/femap.py
```diff
@@ -346,20 +392,34 @@ def add_absolute_calculation(
...
         Note
         ----
         The pandas DataFrame will have the following columns:
+
         - labelA
         - labelB
-        - DDG (kcal/mol)
-        - uncertainty (kcal/mol)
+        - ``DDG (kcal/mol)`` / ``ΔpIC50`` — depending on ``observable_type``
+        - ``uncertainty (kcal/mol)`` / ``uncertainty (ΔpIC50)``
```
IAlibay: The ΔpIC50 here is a bit confusing given that the alternative for ΔΔG is a unit. Would "unitless" be any better or more confusing?

cinnabar/femap.py
IAlibay, line 645: This early return doesn't account for the observable unit type requested, so you'll get an empty dataframe with the wrong units if you requested pIC50.

cinnabar/conversion.py
IAlibay: <s>It's not immediately clear to me what this means - what are your expected input types for uncertainty now? Is there a test currently doing this somewhere?</s> Ignore me here - I got confused between your github comment and the actual code comment.

Also this change expects uncertainty to have a `.m`, and therefore to be `unit.Quantity | None`. Whilst this matches the function signature, line 117 historically seems to have been ok with pIC50 values being passed as floats instead. This is a change in expected behaviour, is that what we want to do?

IAlibay (changes requested):
Initial round of reviews, I'm still looking through things.

cinnabar/femap.py
```diff
@@ -22,13 +22,59 @@
...
+        Temperature for the conversion (passed to
+        :func:`~cinnabar.conversion.convert_observable`).
+
+    Returns
+    -------
+    pd.DataFrame
+        Copy of the dataframe with ``value_col`` / ``uncertainty_col`` dropped and the two
+        new columns in their place.
+    """
+    # default units should always be kcal/mol in the FEMap
```
IAlibay, line 68: Is it worth cchecking that `value_col` and `uncertainty_col` are in the column names?

cinnabar/femap.py
```diff
@@ -22,13 +22,59 @@
...
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Input dataframe.  ``value_col`` and ``uncertainty_col`` must be present and
+        contain plain floats in kcal/mol.
+    value_col, uncertainty_col : str
+        Names of the columns to convert.
+    new_value_col, new_uncertainty_col : str
+        Column names to use in the returned dataframe.
```
IAlibay, line 54: missing entries for uncertainty and new_uncertainty

cinnabar/femap.py
```diff
@@ -22,13 +22,59 @@
...
+    pd.DataFrame
+        Copy of the dataframe with ``value_col`` / ``uncertainty_col`` dropped and the two
+        new columns in their place.
+    """
+    # default units should always be kcal/mol in the FEMap
+    values = df[value_col].to_numpy() * _kcalpm
+    uncertainties = df[uncertainty_col].to_numpy() * _kcalpm
+    converted_values, converted_uncertainties = convert_observable(values, "dg", "pic50", uncertainties, temperature)
+    # make sure we do not change the column order the dataframe should feel the same
+    col_order = [
```
IAlibay: This one is on the edge of what's understandable in a single lilne. Is there any kind of multi-line formatting that could make it easier for the next person?

cinnabar/femap.py
```diff
@@ -346,20 +392,34 @@ def add_absolute_calculation(
...
+    def get_relative_dataframe(
+        self,
+        observable_type: ANALYSIS_UNITS = "dg",
+        temperature: Quantity = 298.15 * unit.kelvin,
+    ) -> pd.DataFrame:
         """Get a dataframe of all relative results for all sources including experimental and computational.
 
+        Parameters
+        ----------
+        observable_type : ANALYSIS_UNITS, optional
```
IAlibay: I'm not super fond of doing this, because ANALYSIS_UNITS isn't exposed to users looking at the docstring. Given that we're only on two units, I would prefer if you were just explicit with the `Literal['dg', 'pic50']` and then we deal with the headache of providing more conversions later?

cinnabar/femap.py
```diff
@@ -346,20 +392,34 @@ def add_absolute_calculation(
...
+            The observable type to report values in.  Defaults to ``dg`` (kcal/mol).
+            Use ``pic50`` to report ΔpIC50 values.
+        temperature : Quantity, optional
```
IAlibay: This file mixes `optional` as either "could be None" (see `units` in `read_csv` for an example) or "has a default value" (here).

This makes it rather confusing to the users. Either here, or in a separate PR, could you clean up the docstrings so that we have consistency? My take is a) use `default X` instead of "optional", and b) use `X | None` instead of optional.

cinnabar/femap.py
```diff
@@ -368,43 +428,72 @@ def get_relative_dataframe(self) -> pd.DataFrame:
         for l1, l2, d in self._graph.edges(data=True):
             if d["source"] == "reverse":
                 continue
-            if isinstance(l1, ReferenceState) or isinstance(l2, ReferenceState) and not d["computational"]:
+            # grab only the experimental non-computational results
+            if (isinstance(l1, ReferenceState) or isinstance(l2, ReferenceState)) and not d["computational"]:
```
IAlibay, line 460: Are these tested anywhere?

IAlibay (commented):
Last set of comments for now.

cinnabar/femap.py
jthorton: I agree changed to `DpIC50`!

cinnabar/femap.py
jthorton: Not sure tbh I have changed to unitless for now but I might remove it.

cinnabar/femap.py
jthorton: Added a check!

cinnabar/femap.py
jthorton: Good point I was hoping some magic would happen and the API docs would resolve the type as it does on the function signature but that doesn't seem to be the case, now made each doc string explicit.

cinnabar/conversion.py
jthorton: Yes, originally I was trying to work around users having to pass in pic50 values with `unit.dimensionless` which might be awkward but this isn't inline with what the function expects or how it interacts with other observables, so I would consider this a bug fix and this is now the correct behaviour.

cinnabar/tests/test_conversion.py
jthorton: This should have been done originally as the conversion function expects the values to have units even if that unit is `unit.dimensionless`.

cinnabar/femap.py
jthorton: Ahh I used the numpy convention where they share the type and definition as they are similar 
```
value_col, uncertainty_col : str
    Names of the columns to convert.
new_value_col, new_uncertainty_col : str
    Column names to use in the returned dataframe.
```

cinnabar/femap.py
jthorton: Tried to clean it up lets see what black does

cinnabar/femap.py
jthorton: Thanks, I have done this in #209 so lets merge that first and then I can get any missing clean up here.

cinnabar/femap.py
jthorton: This was originally fixed in #206 and should be tested there.

cinnabar/femap.py
jthorton: Good catch this is correctly handled now and tested!

cinnabar/femap.py
jthorton: Good idea if we ever change the column names or want to use reuse the function a clear error would be better, added a check and a test.

jthorton:
pre-commit.ci autofix

cinnabar/femap.py
jthorton: No it changed it back ...

cinnabar/femap.py
IAlibay: Sorry that's my bad for misreading.

IAlibay (approved):
lgtm, thanks!

Issue #205, opened 2026-05-20
`scatter_kwargs` should reflect the `default_kwargs` values


IAlibay:
Just looking at the `_master_plot` (or now `pair_plot`) code, I noticed that the default_kwargs don't match scatter_kwargs, which makes the choice of defaults obscure to users. The default kwarg dictionary should have all the default values.

jthorton:
Fixed in #209

PR #209, opened 2026-05-22
add mypy to CI

### LINKED ISSUE #207: Add mypy checks to CI
Probably should do this as part of the usability drives.

jthorton:
## Description
Fixes #207 #205 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready to go

Tips


jthorton:
pre-commit.ci autofix

jthorton:
pre-commit.ci autofix

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+  mypy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
IAlibay:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - "**.py"
+
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash -l {0}
```
IAlibay:
```suggestion
    shell: bash -leo pipefail {0}
```
Generally Mike recommends this as the more robust way of doing it.

cinnabar/conversion.py
```diff
@@ -114,19 +114,19 @@ def convert_observable(
 
     # if the input is in pic50, add dimensionless units for consistent conversions
     if original_type == "pic50":
```
IAlibay: This bit confuses me a little bit, it seems to be indicating that you could pass in `value` as a float and it'll convert it to a Quantity, but then above & in the tests, you say it should always be a Quantity.

I think this is related to similar review comments in #208

cinnabar/femap.py
```diff
@@ -12,12 +12,11 @@
 import pathlib
 import warnings
 from dataclasses import asdict
-from typing import TYPE_CHECKING, Hashable, Optional, Union
+from typing import TYPE_CHECKING, Hashable, Optional
```
IAlibay: You're moving a bunch of things from `Optional` to `X | None` (which is great!), is `Optional` still required as an import?

cinnabar/femap.py
```diff
@@ -30,14 +29,14 @@
...
-def read_csv(filepath: pathlib.Path, units: Optional[openff.units.Quantity] = None) -> dict:
+def read_csv(filepath: pathlib.Path, units: Quantity | None = None) -> dict:
     """Read a legacy format csv file
 
     Parameters
     ----------
     filepath
       path to the csv file
-    units : openff.units.Quantity, optional
+    units : Quantity, optional
```
IAlibay: As mentioned in #208 the term `optional` is used very loosely in cinnabar docstrings. Might be worth being more explicit as `Quantity | None, default None`?

cinnabar/femap.py
```diff
@@ -30,14 +29,14 @@
 _kcalpm = unit.kilocalorie_per_mole
 
 
-def read_csv(filepath: pathlib.Path, units: Optional[openff.units.Quantity] = None) -> dict:
+def read_csv(filepath: pathlib.Path, units: Quantity | None = None) -> dict:
```
IAlibay:
```suggestion
def read_csv(filepath: pathlib.Path, units: Quantity | None = None) -> dict[str, dict[Hashable, Measurement] | list[[Measurement]]
```
Probably too much, but just in case you think it's worth it.

cinnabar/femap.py
```diff
@@ -244,17 +244,17 @@ def add_experimental_measurement(
         ----------
         label
```
IAlibay: Add typing to docstring?

cinnabar/femap.py
```diff
@@ -244,17 +244,17 @@ def add_experimental_measurement(
...
           the ligand being measured
-        value : openff.units.Quantity
+        value : Quantity
```
IAlibay: Here and everywhere else - sorry to ask but can you revert back to `openff.units.Quantity`? We try to keep it as `openff.units.Quantity` in docstring, even though it's imported as `Quantity`. This is because the openff.units.Quantity doesn't necessarily match the pint.Quantity and then there's the whole openmm Quantity thing.

You can see examples of this in `openfe`.

cinnabar/measurements.py
```diff
@@ -89,26 +89,26 @@ class Measurement:
...
-        uncertainty: unit.Quantity | str,
+        DG: Quantity | str,
+        uncertainty: Quantity | str,
         computational: bool,
         source: str = "",
-        temperature: unit.Quantity | str = 298.15 * unit.kelvin,
+        temperature: Quantity | str = 298.15 * unit.kelvin,
     ):
         """
         Initialize a Measurement object converting all quantities to the correct default units.
```
IAlibay, line 114: I know it was missing historically, but whilst we're here - could you add in a `Parameters` section to this docstring?

cinnabar/plotting.py
```diff
@@ -160,9 +160,12 @@ def _master_plot(
     cm = plt.get_cmap("coolwarm")
 
     if color is None:
-        color = np.abs(x - y)
+        _color = np.abs(x - y)
```
IAlibay, line 167: That's a bit of an odd one to have to fix - out of curiosity what was mypy complaining about?

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,47 @@
+name: "mypy"
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+    paths:
+      - "**.py"
```
IAlibay, line 12: Just checking that it's intentional to skip non-python files changes in push but not in pull_request

IAlibay (changes requested):
Mostly small things, otherwise lgtm.

jthorton:
pre-commit.ci autofix

cinnabar/femap.py
jthorton: Yeah no problem, not sure why mypy is not keen to have that be the type hint that would make things easier, but reverted.

cinnabar/femap.py
jthorton: Thanks I think this is the best place to do it, changed everywhere.

.github/workflows/mypy.yaml
jthorton: Ahh good catch no that was a mistake!

cinnabar/conversion.py
jthorton: Yeah, this was a mistake in the unit conversion PR. Everything should have units, but adding `unit.dimensionless` to pic50 values for users may not feel natural, so I worked around it, and mypy is flagging it as an issue as well. 

I planned to remove this. I have now done this here as this will probably merge first.

cinnabar/femap.py
jthorton: The only thing left over is the forward reference to the estimator, which needs the `Optional` type I could try and remove it, though.

cinnabar/measurements.py
jthorton: Done.

jthorton:
pre-commit.ci autofix

cinnabar/femap.py
jthorton: Done.

cinnabar/plotting.py
jthorton: mypy was complaining that the type was still a string or None but using a different variable name got around this, everything else I tried had an error :/

jthorton:
pre-commit.ci autofix

cinnabar/plotting.py
IAlibay: How odd - this would have been one of those cases where telling mypy to ignore the line would have been a reasonable solution too :)

cinnabar/stats.py
```diff
@@ -486,21 +486,27 @@ def mle(graph: nx.DiGraph, factor: str = "f_ij", node_factor: Union[str, None] =
     return f_i, C
 
 
-def form_edge_matrix(graph: nx.Graph, label: str, step=None, action=None, node_label=None) -> np.ndarray:
+def form_edge_matrix(
+    graph: nx.Graph,
+    label: str,
+    step=None,
```
IAlibay, line 492: Is step just a dead argument? If so, could you (maybe in a separate PR) deprecate it for removal in the next release?

We could just remove it here, but for the sake of API stability it might be worth a deprecation cycle in case.

IAlibay (approved):
Lgtm! Just the one thing, probably as a follow-up issue/PR.

cinnabar/stats.py
jthorton: Yeah it seems dead, the plan is to remove these functions completely in Ians work on updating the MLE estimator, maybe I should start the deprecation cycle now for the function in another PR?

PR #231, opened 2026-07-13
Small corrections to one of the example guides.

dotsdl:


examples/cinnabar-api.ipynb
```diff
@@ -176,9 +176,9 @@
    "id": "5ba98e410bf98178",
    "metadata": {},
    "source": [
-    "A key property to consider is whether the graph is fully connected, that is, whether all results are reachable from other results. The ``FEMap`` object provides a method to check this and a visualisation method which can be helpful to identify missing connections in the network.\n",
+    "A key property to consider is whether the graph is connected, that is, whether all results are reachable from other results. The ``FEMap`` object provides a method to check this and a visualisation method which can be helpful to identify missing connections in the network.\n",
```
jthorton: Thanks thats a good catch, fully connected is wrong! What about changing this to weakly connected as that is what is needed and would align with the API point `check_weakly_connected`?

jthorton (approved):
LGTM thanks @dotsdl!

PR #240, opened 2026-09-18
Fix CI - duecredit

jthorton:
## Description

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## LLM / AI generated code disclosure

LLMs or other AI-powered tools (beyond simple IDE use cases) were used in this contribution: yes / no
If yes, please provide details here:

## Checklist
- [ ] Filled in the AI-generated code disclosure.

## Status
- [x] Ready to go

Tips


jthorton:
This doesn't seem to fix the 3.13 runs but it does fix the warnings in the other runs they now print the citation correctly!

jthorton:
pre-commit.ci autofix

devtools/conda-envs/env.yaml
```diff
@@ -13,7 +13,8 @@ dependencies:
   - scikit-learn
   - adjusttext
   - openff-units
-  - duecredit<0.10
```
jthorton, line 16: Both are needed to make sure the duecredit report works!

devtools/conda-envs/env.yaml
```diff
@@ -13,7 +13,8 @@ dependencies:
...
+  - duecredit
```
atravitz, line 16: can we min-pin this since you know it was failing with a lower version?

devtools/conda-envs/env.yaml
```diff
@@ -13,7 +13,8 @@ dependencies:
...
+  - citeproc-py-styles  # needed for duecredit harvard1 style
```
atravitz, line 17: why do we need this in cinnabar but not openfe? I'd like to not pull in a dep unless we _need_ to.

jthorton: It looks like this is [broken in openfe too](https://github.com/OpenFreeEnergy/openfe/actions/runs/35584285306/job/106283902461#step:11:4100) and we might be able to pin `citeproc-py` to fix this.

jthorton:
```suggestion
  - citeproc-py <0.10.3  # the default harvard1 style used by duecredit was changed in 0.10.3, this avoids needing an extra dependency on citeproc-py-styles 
```

jthorton: Upper pinning citeproc-py seems to have fixed it!

jthorton:
pre-commit.ci autofix

devtools/conda-envs/env.yaml
jthorton: I think this removes the need for the lower pin?

atravitz: ah got it, thanks!

atravitz (approved):
thanks for tracking this down @jthorton! I think we could consider changing the citation style to something supported natively by citeproc-py, but I haven't looked into that yet.

This'll do for now!

